PR #1, opened 2016-01-10
Add Design for Home Page

dak:
- [x] Update Header
 - [x] Update Footer
 - [x] Add static design for US History banner
 - [x] Make Header transition on scroll
 - [x] Add responsive support for all desktop browser widths
 - [x] Verify WCAG A support

Because this PR is already quite large, the remaining tasks will be completed in a subsequent PR.
 - Add quote section
 - Add 2nd hero banner
 - Add 3 bottom banners
 - Add all book banners and intelligently load/pre-fetch them
 - Create parallax view and add parallax to banner elements
 - Add missing resources
 - Add ☰ to collapsed header and relevant supporting functionality

**Note:** This PR will only add the US History hero banner. Additional design info is required to determine how best to structure the code to transition and support multiple banners, as well as to transition between them. Current thought is they should all be separate modules within the home page directory that are lazy-loaded when needed (with pre-fetching).

**Note:** WCAG AA Support is currently falling short due to a lack of contrast.  The orange "Explore U.S. History" button and all text in the footer are a few examples of contrast checks that are currently failing.  This is a design issue that can't be fixed in development, and as such only WCAG A will be targeted in development.

**Note:** Resources are needed for all social icons (ideally as SVG). The Subject Card also needs to be properly exported as an SVG as this PR is currently using a hack to size it properly.

image

image

image

image


src/app/components/shell/footer/footer.hbs
```diff
@@ -1,64 +1,62 @@
...
-        <nav>
-            <ul>
-                <li><a href="/license">Licensing</a></li>
-                <li><a href="/tos">Terms of Use</a></li>
-                <li><a href="/contact">Contact</a></li>
-            </ul>
-        </nav>
+        <ul>
+            <li>
+                <a href="/licensing">Licensing</a>
```
RoyEJohnson: This is weird. A list with one item that contains three distinct things? Is there a reason not to make this just a div with a class?

src/app/components/shell/footer/footer.hbs
```diff
@@ -1,64 +1,62 @@
...
 
-    <div class="connect">
-        <span class="small-header">Connect</span>
-        <ul>
-            <li><a href="https://play.google.com/store/apps/details?id=org.cnx.android">Android App</a></li>
-            <li><a href="//devblog.cnx.org">Dev Blog</a></li>
-            <li><a href="http://itunes.apple.com/us/institution/connexions/id375198208">iTunes U</a></li>
-        </ul>
+        <div>
+            Advanced Placement® and AP® are trademarks registered and/or
```
RoyEJohnson: Probably want `&reg;` here, similar to how you replaced © with `&copy;` above.

src/app/components/shell/footer/footer.hbs
dak: Agreed, that should have been three separate list items. Fixed.

src/app/components/shell/footer/footer.hbs
dak: Fixed.

src/app/components/shell/header/header.js
```diff
@@ -1,21 +1,88 @@
...
+    isPinned() {
+        return !!this.classList('contains', 'fixed');
+    }
+
+    isTransparent() {
+        return !!this.classList('contains', 'transparent');
+    }
+
+    get height() {
+        let header = this.el.querySelector('.page-header');
```
RoyEJohnson: Q: I've been doing one `let` with a comma-separated list of declarations; you use one `let` per declaration. Do we want to standardize, or does it not matter?

dak: I think the original reasoning for a single `var` in a function scope was because `var` is hoisted to the top of the scope anyway, so it helps to prevent errors.  Since `let` is a bit more flexible and can be used in block scope, I don't think that's as relevant as a best practice any more.

In cases like that, I tend to defer to the most common practices in other languages, which use multiple declaration statements.  I don't think it has much impact on code clarity either way though, so I'm not opposed to leaving it up to the coder to decide what is best in a given situation (we could enforce the rule either way in ESLint, but I left it disabled at the moment at least to allow the choice).

src/app/pages/home/home.js
```diff
@@ -11,7 +12,24 @@ import {template} from './home.hbs';
 export default class Home extends BaseView {
 
     onRender() {
-        // this.regions.featured.show(new CarouselView());
+        this.updateHeaderStyle();
+        window.addEventListener('scroll', this.updateHeaderStyle.bind(this));
+    }
+
+    updateHeaderStyle() {
+        if (window.pageYOffset > 50 && !appView.header.isPinned()) {
```
RoyEJohnson: Should probably have some variables instead of magic numbers 50 and 10.

dak: The secondary nav height is now dynamically calculated.  `10` is just the conversion from `px` to `rem` which is defined for this project.

src/styles/components/responsive.scss
```diff
@@ -12,48 +14,48 @@
...
     @include phone {
-        display: none;
+        display: none !important;
     }
 }
 
 .visible-desktop {
     @include phone {
-        display: none;
+        display: none !important;
```
RoyEJohnson: You have `.visible-desktop` defined exactly the same as `.hidden-desktop`, with `display: none`. Is that right? (Similarly the other visibles and hiddens.)

PR #3, opened 2016-01-12
Add License Page

dak:


src/app/pages/license/license.hbs
```diff
@@ -0,0 +1,206 @@
...
+                in our repository.
+            </p>
+
+            <h3>How do I control what changes are made to my modules?</h3>
+            <p>
+                Only persons who have the maintainer role on your module have permission
+                to modify and publish an updated version of your module. As the author
+                of a module, you determine who has the maintainer role on your module.
+                The Creative Commons Attribution license does allow another person to
+                make an <a href="../authoring/expressedit">adaptation</a> of any content
```
edwoodward: This is a link to help in Legacy. It should be removed.

src/app/pages/license/license.hbs
```diff
@@ -0,0 +1,206 @@
...
+                clearly visible and clearly explains the uses people may legally make
+                of your work. Strictly speaking, fraud and plagiarism are not the subject
+                of copyright law, but copyright law is a good first defense.
+            </p>
+
+            <h3>Can people make money off of my work?</h3>
+            <p>
+                It is possible, but to be effective at all the license we use must allow
+                some "commercial" uses. Consider the situation in which a printing service,
+                such as FedEx Kinko's℠, offers to print your work so that your students
```
edwoodward: I think this should be changed to "local printing service". The Kinko's name is no longer used.

src/app/pages/license/license.hbs
```diff
@@ -0,0 +1,206 @@
...
+                that you include outside of our repository. This attribution is as simple
+                as clearly naming the original author and identifying the portion of
+                his or her work you are using. In addition you must retain the copyright
+                notice and provide the URI of the original work.
+            </p>
+
+            <h3>What must I do if I want to include another author’s module in a collection
+            that I am building?</h3>
+            <p>
+                Simply add the module to your collection with the Collection Composer tool.
```
edwoodward: This should be: "Simply add the page to your book." Remove the reference to the collection composer.

src/styles/main.scss
```diff
@@ -6,6 +6,7 @@ html {
     font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
     font-size: 62.5%;
     font-weight: 300;
+    line-height: 1.42857143;
```
RoyEJohnson: How do you arrive at this highly-precise number?

dak: It's 20/14.  It's a commonly used number by other frameworks as well, including Bootstrap.

RoyEJohnson: Might be clearer to use SASS's math ability to just make it 20 / 14. But I'm gonna merge.

philschatz: Why is this missing the units?

dak: No units is allowed for line-height.  From the spec, if you only include a unitless number it `will be multiplied with the current font size to set the line height`.

That's actually what we want, since it allows it to scale properly with font-size changes.

philschatz: ah, it's multiplied by the element's font size :smile:

PR #6, opened 2016-01-19
Add Adoption Form page

RoyEJohnson:
Fields validate
Form submits to /adoptions, which does not do anything
Next PR is based off of this one

PR #15, opened 2016-02-02
Create interest form

RoyEJohnson:
Moved form css into shared resource
Put hint text for fieldsets into legend

### DIFF
--- a/src/app/pages/interest-form/interest-form.hbs
+++ b/src/app/pages/interest-form/interest-form.hbs
@@ -1,0 +1,123 @@
+<h1>OpenStax College Interest Form</h1>
+<p>
+    Thank you for your interest in OpenStax College textbooks! Please complete the form below so that we may provide
+    you with the information you need to start saving your students thousands of dollars each semester.
+</p>
+<form accept-charset="UTF-8" action="/adoption_interests" class="new_adoption_interest" method="post">
+    <p>Have you already adopted an OpenStax College textbook? <a href="/adoptions">Let us know!</a></p>
+    <h2>Contact Information</h2>
+    <fieldset class="labeled-inputs">
+        <label>
+            <span>First name</span>
+            <input name="adoption_interest[first_name]" type="text" required />
+        </label>
+        <label>
+            <span>Last name</span>
+            <input name="adoption_interest[last_name]" type="text" required />
+        </label>
+        <label>
+            <span>School name</span>
+            <input name="adoption_interest[school_name]" type="text" required />
+        </label>
+        <label>
+            <span>School url</span>
+            <input name="adoption_interest[school_url]" type="text" required />
+        </label>
+        <label>
+            <span>Email address</span>
+            <input name="adoption_interest[email_address]" type="email"
+             pattern="[^@]+@([^@\.]+\.)+[^@\.]+" required />
+        </label>
+        <label>
+            <span>Phone number</span>
+            <input name="adoption_interest[phone_number]" type="text" required />
+        </label>
+    </fieldset>
+    <h2>Textbook Interest</h2>
+    <p>OpenStax College textbook are you interested in adopting:</p>
+    <fieldset class="labeled-buttons require-one">
+        <legend class="validation-message">
+            Select at least one
+        </legend>
+        {{#each titles}}
+            <label>
+                <input name="adoption_interest[textbook_interests][]" type="checkbox" value="{{this}}">
+                {{this}}
+            </label>
+        {{/each}}
+    </fieldset>
+    <fieldset class="labeled-inputs">
+        <label>
+            <span>Students per semester</span>
+            <input name="adoption_interest[students_per_semester]" type="number" required>
+        </label>
+    </fieldset>
+    <p>
+        Which partners may we share your information with so they may contact you about
+        additional resources supporting the OpenStax College books?
+    </p>
+    <fieldset class="labeled-buttons">
+        <label>
+            <input name="adoption_interest[partner_resource][]" type="checkbox" value="Online homework providers">
+            Online homework providers
+        </label>
+        <label>
+            <input name="adoption_interest[partner_resource][]" type="checkbox" value="Courseware providers">
+            Courseware providers
+        </label>
+        <label>
+            <input name="adoption_interest[partner_resource][]" type="checkbox" value="Print custom editions">
+            Print custom editions
+        </label>
+    </fieldset>
+    <p class="">How did you hear about OpenStax College?</p>
+    <fieldset class="labeled-buttons require-one">
+        <legend class="validation-message">
+            Select at least one
+        </legend>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Web search">
+            Web search
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Colleague">
+            Colleague
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Conference">
+            Conference
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Email">
+            Email
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Facebook">
+            Facebook
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Twitter">
+            Twitter
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Webinar">
+            Webinar
+        </label>
+        <label>
+            <input name="adoption_interest[referral][]" type="checkbox" value="Partner organization">
+            Partner organization
+        </label>
+        <label>
+            <input class="other" name="adoption_interest[referral][]" type="checkbox" value="">
+            Other:
+            <input  name="adoption_interest[other]" type="text">
+        </label>
+    </fieldset>
+    <div>
+        <label>
+            <p>Any feedback you'd like to share with us?</p>
+            <textarea cols="50" name="adoption_interest[feedback]" rows="6"></textarea>
+        </label>
+    </div>
+    <input name="commit" type="submit" value="Submit" role="button" aria-disabled="false" />
+</form>
--- a/src/app/pages/interest-form/interest-form.js
+++ b/src/app/pages/interest-form/interest-form.js
@@ -1,0 +1,64 @@
+import BaseView from '~/helpers/backbone/view';
+import {on, props} from '~/helpers/backbone/decorators';
+import {template} from './interest-form.hbs';
+
+const otherBoxSelector = '.other[type="checkbox"]',
+    otherTextSelector = 'input[name$="[other]"]';
+
+let bookTitles = [
+    'Algebra and Trigonometry',
+    'Anatomy and Physiology',
+    'Biology',
+    'Chemistry',
+    'College Algebra',
+    'College Physics',
+    'College Physics for AP® Courses',
+    'Concepts of Biology',
+    'Introduction to Sociology',
+    'Introduction to Sociology 2e',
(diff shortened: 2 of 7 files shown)


src/app/pages/interest-form/interest-form.scss
```diff
@@ -0,0 +1 @@
+@import 'forms';
```
dak: Can you move the `forms.scss` file into `/components/forms.scss` and import at the bottom of `main.scss` (like buttons, headers, etc)?  This file can then be deleted if no unique styling is necessary for this page.

src/app/router.js
```diff
@@ -39,6 +39,10 @@ class Router extends Backbone.Router {
         this.route('adoptions', 'adoption-form', () => {
             shell.render('adoption-form');
         });
+
+        this.route('adoption_interests', 'interest-form', () => {
```
dak: `/adoption_interests` seems like an unintuitive URL.  Would something like just `/interest` or `/adopt` make more sense?

src/styles/forms.scss
```diff
@@ -0,0 +1,49 @@
...
+    background-color: color(primary);
+    color: white;
+    font-size: 150%;
+}
+
+fieldset {
+    &.invalid,
+    input:invalid,
+    select:invalid {
+        box-shadow: 0 0 5px color(danger);
```
dak: px -> rem

src/styles/forms.scss
```diff
@@ -0,0 +1,49 @@
+@import 'variables';
+
+[type="submit"] {
+    background-color: color(primary);
+    color: white;
+    font-size: 150%;
```
dak: If you want a specific font size, can you use `rem` instead of `%`?

src/styles/forms.scss
```diff
@@ -0,0 +1,49 @@
...
+            display: block;
+        }
+    }
+
+    > .validation-message {
+        display: none;
+    }
+
+    &.invalid > .validation-message {
+        background-color: white;
```
dak: Try to use variables for colors, even black and white.  `white` is `color(neutral-lightest)`.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
+<style scoped>@import url(/app/pages/interest-form/interest-form.css);</style>
+<h1>OpenStax College Interest Form</h1>
+<p>
+    Thank you for your interest in OpenStax College textbooks! Please complete the form below so that we may provide
+    you with the information you need to start saving your students thousands of dollars each semester.
+</p>
+<form accept-charset="UTF-8" action="/adoption_interests" class="new_adoption_interest" method="post">
+    <div style="margin:0;padding:0;display:inline">
```
dak: If you need some styling, it should be moved into the SCSS file for this page.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+        <input name="utf8" type="hidden" value="✓">
+        <input name="authenticity_token" type="hidden" value="Xi14qRzCbSy2zkso1VEjKU4bEUfFWOrhWbQnh7tlO3c=">
```
dak: I don't think this should be a fixed value.  I'm assuming this is for CSRF.  It should be a random value provided by the server.  I believe Rails happens to call it an `authenticity_token`, although you could name it whatever makes sense.  You may need to talk to the backend though to see if they provide anything for this, whoever happens to be processing the form (is it SalesForce?).

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+            <input name="adoption_interest[school_name]" type="text" required="true" />
+        </label>
+        <label>
+            <span>School url</span>
+            <input name="adoption_interest[school_url]" type="text" required="true" />
+        </label>
+        <label>
+            <span>Email address</span>
+            <input name="adoption_interest[email_address]" type="email"
+             pattern="[^@]+@([^@\.]+\.)+[^@\.]+" required="true" />
```
dak: Don't use `required="true"`.  `required="false"` also is `true`.  For XHTML you use `required="required"`, although since we're just plan HTML5, you can just truncate it to the property itself: `required`.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="College Physics for AP® Courses ">
+            College Physics for AP® Courses
+        </label>
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Concepts of Biology">
+            Concepts of Biology
+        </label>
+        <label>
+            <input id="adoption_interest_textbook_interests_" name="adoption_interest[textbook_interests][]" type="checkbox" value="Introduction to Sociology">
```
dak: A number of elements have identical ids, which is not valid HTML.  For example, I'm seeing 3 different elements with `adoption_interest_textbook_interests_`.  It doesn't appear this ids are being used for anything either, in which case they should just be removed.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+        <label>
+            <input name="adoption_interest[partner_resource][]" type="checkbox" value="Courseware providers">
+            Courseware providers
+        </label>
+        <label>
+            <input name="adoption_interest[partner_resource][]" type="checkbox" value="Print custom editions">
+            Print custom editions
+        </label>
+    </fieldset>
+    <p class="">How did you hear about OpenStax College?
```
dak: Where's the closing tag for this element?  And it has an empty class, which should just be deleted.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+            U.S. History
+        </label>
+    </fieldset>
+    <fieldset class="labeled-inputs">
+        <label>
+            <span>Students per semester</span>
+            <input name="adoption_interest[students_per_semester]" type="number" required="true">
+        </label>
+    </fieldset>
+    <div class="section">
```
dak: This element appears to have no closing tag and doesn't seem to be necessary.  It can probably just be deleted.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+    <fieldset class="labeled-buttons require-one">
+        <legend class="validation-message">
+            Select at least one
+        </legend>
+        <label>
+            <input id="adoption_interest_referral_" name="adoption_interest[referral][]" type="checkbox" value="Web search">
+            Web search
+        </label>
+        <label>
+            <input id="adoption_interest_referral_" name="adoption_interest[referral][]" type="checkbox" value="Colleague ">
```
dak: Is the value of this input intended to have an empty space at the end of it?

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,203 @@
...
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Principles of Macroeconomics for AP® Courses">
+            Principles of Macroeconomics for AP® Courses
+        </label>
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Principles of Microeconomics">
+            Principles of Microeconomics
+        </label>
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Principles of Microeconomics for AP® Courses">
+            Principles of Microeconomics for AP® Courses
```
dak: HTML code for ® is `&reg;`.

src/app/pages/adoption-form/adoption-form.js
```diff
@@ -71,5 +71,4 @@ export default class AdoptionForm extends BaseView {
         this.regions.primaryBook.append(bookInfoView);
         this.toggler.update();
     }
-
```
dak: I think this blank line was actually intended.  We seem to be leaving a blank line at the start and end of classes.

dak:
Can you rebase these into a single commit?  Also, could you change the commit messages to follow the general guidelines for a good git commit message found [here](http://chris.beams.io/posts/git-commit/) among a number of other places?

image

src/app/pages/adoption-form/adoption-form.js
RoyEJohnson: The file still has a blank line at the bottom in my editor, and ends on line 75 (close-brace is on 74). ???

dak: There's an empty line at the start and end of every other class, I believe.  For example, line 44 is blank at the opening of the class:
```
export default class AdoptionForm extends BaseView {

    @on('click .toggle-section')
```

I think you also want a blank line between the `}`s on line 73 and 74 (which did exist, but was removed here).

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,201 @@
+<h1>OpenStax College Interest Form</h1>
+<p>
+    Thank you for your interest in OpenStax College textbooks! Please complete the form below so that we may provide
+    you with the information you need to start saving your students thousands of dollars each semester.
+</p>
+<form accept-charset="UTF-8" action="/adoption_interests" class="new_adoption_interest" method="post">
+    <div>
+        <input name="utf8" type="hidden" value="✓">
+        <input name="authenticity_token" type="hidden" value="Xi14qRzCbSy2zkso1VEjKU4bEUfFWOrhWbQnh7tlO3c=">
```
dak: This appears to be a CSRF token generated by rails, probably on the form you were duplicating.  It wouldn't have any meaning here, and would need to be randomly generated (and verified) by the server.

I'm all for using a CSRF token, but we need to have server support to do so.  So I think we either need to coordinate with @rich-hart and @mwharrison or just remove this line for the time being.

It looks like this also [exists on the adoption form page](https://github.com/openstax/os-webview/pull/6/files#diff-985278373ff40a89fe057052763be39fR18), although I missed it before it got merged.

src/app/pages/interest-form/interest-form.hbs
```diff
@@ -0,0 +1,201 @@
...
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Principles of Microeconomics for AP® Courses">
+            Principles of Microeconomics for AP&reg; Courses
+        </label>
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="Psychology">
+            Psychology
+        </label>
+        <label>
+            <input name="adoption_interest[textbook_interests][]" type="checkbox" value="U.S. History">
+            U.S. History
```
dak: I wonder if, rather than manually listing all of these books, it wouldn't be better to create a model representing the books and then just use an `{{#each books}}` loop in the template to generate them.

src/app/pages/interest-form/interest-form.js
```diff
@@ -0,0 +1,74 @@
...
+    otherTextSelector = 'input[name$="[other]"]';
+
+function bookInfoDict(label, value) {
+    if (!value) {
+        value = label;
+    }
+    return {
+        label,
+        value
+    }
```
dak: You could write this as:
```javascript
return {
    label,
    value || label
};
```
to fix two of the ESLint errors.  The third error appears to be that this function is never actually used anywhere, so if it's dead code I guess just remove it and everything should be fixed.

src/app/pages/interest-form/interest-form.hbs
dak: Looking into further, I think you could just remove this entire `<div>` and both hidden values (since we aren't getting a CSRF token from the server at the moment).  The utf8 hidden value exists to trick old versions of IE into properly submitting the form as UTF-8, since it doesn't respect the defined charset, although that issue isn't relevant on any browsers we support.

PR #18, opened 2016-02-04
Updated Navigation Menu

zroehr:
New PR for navigation with most up-to-date master included.

 - [x] Make Header responsive
 - [x] Create mobile interactions
 - [x] Add JS for static header when scrolled
 - [x] Add interactions for viewing/hiding top bar 
 - [x] Update nav links to reflect updated design
 - [x] Code refactor & clean up
 - [x] Cross browser test
 - [ ] Verify WCAG A support.


dak:
This needs to be rebased (and the wip tag removed if it's ready).  It also has a lot of merges and generic commits/messages that should probably be rebased down to a single commit.

src/app/components/shell/header/header.hbs
```diff
@@ -1,41 +1,56 @@
...
...29.3 14 130.1 12.1 130.1 12.1 130.1 14 130.4 14 130.4 11.8 " style="fill:#5E6062;" /><polygon class="st5" points="125.7 12.1 126.4 12.1 126.4 14 126.8 14 126.8 12.1 127.5 12.1 127.5 11.8 125.7 11.8 " style="fill:#5E6062;" /><path id="green" class="st11" d="M47.5 6.5c0 0.8-0.8 1.5-1.6 1.5L9.6 6.3c-0.9 0-1.5-0.8-1.5-1.6l0.2-3.2C8.3 0.6 9.1 0 9.9 0l36.2 1.7c0.9 0 1.5 0.8 1.5 1.6L47.5 6.5z" style="fill:#77AF42;" /><path id="grey" class="st5" d="M35.6 18.2c0 0.5-0.4 0.9-0.9 0.9H5.2c-0.5 0-0.9-0.4-0.9-0.9v-4.6c0-0.5 0.4-0.9 0.9-0.9h29.5c0.5 0 0.9 0.4 0.9 0.9V18.2z" style="fill:#5E6062;" /><path class="st12" d="M44.6 23.1c0 0.5-0.4 0.9-0.9 1l-33.2 1c-0.5 0-0.9-0.4-0.9-0.9l-0.1-2.7c0-0.5 0.4-0.9 0.9-1l33.2-1c0.5 0 0.9 0.4 0.9 0.9L44.6 23.1z" style="fill:#F4D019;" /><path id="blue" class="st36" d="M40.2 28.7c0 0.5-0.4 0.9-0.9 0.9H7.6c-0.5 0-0.9-0.4-0.9-0.9v-1.8c0-0.5 0.4-0.9 0.9-0.9h31.7c0.5 0 0.9 0.4 0.9 0.9V28.7z" style="fill:#00246A;" /><path class="st8" d="M41 10.5c0 0.4-0.6 0.7-1.3 0.7H1.3c-0.7 0-1.3-0.3-1.3-0.7V9c0-0.4 0.6-0.7 1.3-0.7h38.5C40.4 8.3 41 8.6 41 9V10.5z" style="fill:#F47641;" /></g></g><g id="nav"/><g id="Mobile"/><g id="guides" class="st45" style="display:none;" /></svg>
```
dak: I don't think we should be injecting SVGs directly into the HTML.

src/app/components/shell/header/header.hbs
```diff
@@ -1,41 +1,56 @@
...
-        <ul class="nav-menu">
-            <li class="nav-menu-item"><a href="/subjects">Subjects</a></li>
-            <li class="nav-menu-item"><a href="/higher-ed">Higher Education</a></li>
-            <li class="nav-menu-item"><a href="/k-12">K-12</a></li>
-            <li class="nav-menu-item"><a href="/research">Research</a></li>
         </ul>
-    </nav>
+
+          <button class="expand-nav" type="button" role="button" aria-label="Toggle Navigation">
+            <span></span>
```
dak: Should there be an empty span here?

src/app/components/shell/header/header.scss
```diff
@@ -1,12 +1,31 @@
...
+    height: 70px;
+
+    @include medium-up {
+      height: 120px;
+      position: absolute;
+      top: auto;
+      left: auto;
+      right:auto;
+      overflow: visible;
+    }
```
dak: Whitespacing seems to be inconsistent.  There's 4 spaces for indentation beginning with `.page-header`, but 2 spaces in `@include medium-up {`.  Can you make this a consistent 4 spaces (in every file)?

src/app/components/shell/header/header.scss
```diff
@@ -19,6 +38,17 @@
         background-color: transparent;
     }
 
+    //zach class styles
+    &.active {
+      @include small {
+        right: 0;
+        bottom: 0;
+        width:100%;
+        height:100%;
```
dak: There should be a single space between the `:` for a property and its value.

src/app/components/shell/header/header.scss
```diff
@@ -19,6 +38,17 @@
...
+        background: #fff;
```
dak: Try to use the global color variables.  This should be `background: color(neutral-lightest);`.

src/app/components/shell/header/header.scss
```diff
@@ -44,182 +74,583 @@
...
+      .nav-menu-item {
+        &.parent > a {
+          position: relative;
+          &::before,
+          &::after {
+            border-right: 2px solid;
+            content: '';
+            display: block;
+            height: 10px;
+            margin-top: -7px;
```
dak: Avoid using `px`.  You can use `rem` instead (10px == 1rem, so it's an easy conversion).  `rem` is more flexible and allows for better support for accessibility.

src/app/components/shell/header/header.scss
```diff
@@ -44,182 +74,583 @@
...
+            position: absolute;
+            transform: rotate3d(0,0,1,135deg);
+            right: -20px;
+            top: 50%;
+            width: 0;
+          }
+
+          &::after {
+            margin-top: -1px;
+            transform: rotate3d(0,0,1,45deg);
```
dak: Just for consistency, separate comma-separated values with a space.

src/app/components/shell/header/header.scss
```diff
@@ -44,182 +74,583 @@
...
+
+
+
+      } //end nav-menu
+
+      .logo-wrapper {
+        flex: 1 100%;
+        background: transparentize(#E5E5E5, 0.8);
+        padding: 1.5rem;
+        border-top: 1px solid transparentize(#E5E5E5, 0.5);
```
dak: I'm not sure a `transparentize` function is necessary.  CSS's `rgba` is standard and has the exact same effect.  You can use the global variables mixed in with `rgba` as well, so you can do something like `rgba(color(gray), 0.5)`.

src/app/components/shell/header/header.scss
```diff
@@ -44,182 +74,583 @@
...
+                  transform: translate3d(1000px,0,0);
+                  transition: transform;
+                }
+              }
+              .back {
+                @include small {
+                  @at-root .active.open#{&} {
+                    width: 27px;
+                    height: 26px;
+                    text-indent: -9999px;
```
dak: This is an old hack to hide something off screen, but it's still visible to screen readers.  If something should not be visible, can it be down using `display: none` or some other more standard CSS that would also properly convery intent to screen readers?

src/app/pages/home/home.js
```diff
@@ -52,16 +52,12 @@ export default class Home extends BaseView {
...
-            let height = appView.header.height;
-
-            appView.header.reset().collapse().pin();
-            this.el.style.paddingTop = `${height / 10}rem`;
-        } else if (window.pageYOffset <= secondaryNavHeight && !appView.header.isTransparent()) {
-            appView.header.reset().transparent();
-            this.el.style.paddingTop = '0';
+        if (window.pageYOffset >= 250) {
+            appView.header.pin().visible();
+        } else if (window.pageYOffset <= 249 && window.pageYOffset >= 150) {
```
dak: Why make these y offers a fixed value that is dependent on the design (and the users own settings, which can be different! — especially users with poorer vision, who can set a global CSS style to make elements and text larger)?

src/styles/components/buttons.scss
```diff
@@ -22,14 +21,19 @@
...
+    font-weight: 700;
+    padding: 0.5rem 0.75rem;
+}
+
 .btn-large {
     padding: 2.5rem;
 }
 
-@each $name, $type in $os-colors {
+@each $name, $type in $color-types {
```
dak: This should be `os-colors`, not `color-types`.

src/styles/components/buttons.scss
```diff
@@ -22,14 +21,19 @@
...
     .btn-#{$name} {
         background-color: color($name);
-        color: color(neutral-lightest);
+        color: color($name, 'overlay');
```
dak: This line should be reverted.  It looks like it was mistakenly pulled in from one of the merges with one of Roy's old PRs.

src/styles/components/forms.scss
```diff
@@ -1,50 +0,0 @@
-@import 'variables';
-
```
dak: Why is this file being deleted?  Is form styling no longer necessary or being done somewhere else?

src/styles/components/responsive.scss
```diff
@@ -12,50 +12,46 @@
...
-.visible-phone {
-    @include tablet {
-        display: none !important;
-    }
-
-    @include desktop {
+.visible-small {
+    @include medium-up {
         display: none !important;
     }
```
dak: These changes have cascading effects throughout the rest of the site.  Can the names be reverted?

src/styles/main.scss
```diff
@@ -2,6 +2,17 @@
...
+// General overrides
+
+p,
+.btn {
+    margin: 1rem 0 0;
+}
+
+.larger {
+    font-size: 2.4rem;
+}
```
dak: `larger` implies a relative difference, not a fixed size.

src/styles/main.scss
```diff
@@ -12,16 +23,20 @@ html {
 body {
     display: flex;
     flex-direction: column;
-    font-size: 1.4rem;
+    font-size: 1.8rem;
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
```
dak: Why `antialiased` instead of the default or `subpixel-antialiased`?  Subpixel antialiasing looks much better on retina displays, while antialiasing looks better on non-retina.  By default, the OS/browser will choose what it thinks is best.  Usually this only needs to be set on specific elements that are using a scale transform to avoid a pseudo-text-flickering.

src/styles/mixins/responsive.scss
```diff
@@ -1,5 +1,41 @@
...
+    @media (min-width: #{breakpoints(large)}) {
+        @content;
+    }
+}
+
+@mixin large-up { // >= 769
+    @media (min-width: #{breakpoints(medium) + 1}) {
+        @content;
+    }
+}
```
dak: There's so many breakpoints here I think they're losing their usefulness.  I don't know about you, but I don't think I have a good enough memory to keep so many in my head.  They also seem to be a bit arbitrary, and there's some gaps (for example, `medium-down` cover <= 767, while `large-up` covers >= 769, which leaves 768 uncovered in that gap, although it does fall into the range of `medium` which is 641-768, which also overlaps with `medium-down`, etc ....).

How did you come up with these breakpoints?

src/styles/mixins/triangle.scss
```diff
@@ -0,0 +1,52 @@
...
+        border-right: $perpendicular-borders;
+        transform: translateX(-50%);
+
+      } @else if $direction == right or $direction == left {
+        border-bottom: $perpendicular-borders;
+        border-top: $perpendicular-borders;
+        transform: translateY(-50%);
+      }
+    }
+}
```
dak: Is this file part of the nav header changes, or was it pulled in from a merge with Maher's branch at some point?  This is a good example of why it's better to rebase, not merge!

src/styles/variables/colors.scss
```diff
@@ -39,13 +40,34 @@ $neutral-colors: (
...
+        overlay: neutral-color(lightest)
+    ),
+    warning: (
+        color: os-color(yellow),
+        overlay: neutral-color(darkest)
+    ),
+    danger: (
+        color: os-color(red),
+        overlay: neutral-color(lightest)
+    )
```
dak: I think all of these changes are being pulled in from an old merge you did.  It's reverting changes that have been made and this entire file probably needs to be reverted.

src/styles/variables/colors.scss
```diff
@@ -61,7 +83,11 @@ $color-types: (
     }
 
     @if map-has-key($color-types, $color) {
-        @return map-get($color-types, $color);
+        @if $overlay {
+            @return map-get(map-get($color-types, $color), overlay);
+        } @else {
+            @return map-get(map-get($color-types, $color), color);
+        }
```
dak: This also needs to be reverted and looks like it's being pulled in accidentally.

src/styles/variables/opposite-directions.scss
```diff
@@ -0,0 +1,23 @@
...
+        $opposite-direction: map-get($direction-map, $direction);
+        @if $opposite-direction != null {
+		        $opposite-directions: append($opposite-directions, #{$opposite-direction});
+        } @else {
+		      @warn "No opposite direction can be found for `#{$direction}`.";
+		    }
+	  }
+
+    @return $opposite-directions;
+}
```
dak: This file appears to be coming from an unintended merge and should be removed.

src/styles/variables/responsive.scss
```diff
@@ -1,4 +1,7 @@
 $breakpoints: (
+    small: 640px,
+    medium: 768px,
+    large: 1024px,
```
dak: This file probably needs to be reverted.

src/app/components/shell/header/header.hbs
zroehr: this is for the hamburger menu button

src/app/components/shell/header/header.hbs
zroehr: we are going to animate the svg logo so svg needs to be in the HTML

src/app/pages/home/home.js
zroehr: this is for the slide in delay for the header

src/styles/main.scss
melaridi: I did learn something new today. You are right, `subpixel-antialiased` works better on light background with dark font colors. However we would need to add `antialiased` to most areas on homepage i.e: banners 3rd quote (orange section), buckets, footer. 

The font does not render correctly on both retina or non-retina. Here is a demo.

screenshot 2016-02-15 15 33 36

We can have something like this 

`.font-smoothing {
	-webkit-font-smoothing: antialiased;
	-moz-osx-font-smoothing: grayscale;
}`

and add that to areas with white text over colored backgrounds.

src/app/components/shell/header/header.scss
```diff
@@ -19,6 +37,18 @@
         background-color: transparent;
     }
 
+    //zach class styles
```
dak: I'm not sure what `zach class styles` means

src/app/components/shell/header/header.scss
```diff
@@ -1,12 +1,30 @@
+// scss-lint:disable PropertySortOrder SelectorDepth IdSelector NestingDepth
```
dak: These issues should be fixed, not simply told to be ignored

dak:
This is currently causing a number of regressions on the rest of the site, for example:
image
image
image
image

dak:
The breakpoint for switching between the full nav and the hamburger menu appears to be slightly off:
image

src/app/pages/home/home.js
dak: Shouldn't that be dynamically calculated though?  If you want it to slide in below the height of the header, then you could always do something like `appView.header.secondaryNavHeight + 20;`.

src/styles/main.scss
dak: I don't see the issue in that screenshot.  

Yes, the font appears "thicker" or "brighter" between Photoshop and the browser.  It will appear completely different on Windows and Linux.

Be aware that different Operating Systems will render some fonts differently, especially with regard to the algorithms they use for antialiasing and character positioning.  Even different font formats will lead to differences in rendering algorithms, and user settings can also change it:
image
image

(see: https://www.smashingmagazine.com/2012/04/a-closer-look-at-font-rendering/)

That's not something we can (or should try to) control.

zroehr:
the home page regressions you are seeing currently exist on master and are not due to my commits.  I will fix the the navigation error you reported.

dak:
@zroehr https://os-webview-dev.openstax.org is hosting master, and I'm looking at my own branch locally as well.  Those regressions do not exist on master.  You may need to update your master branch:
```
git checkout master
git fetch
git pull
```

PR #19, opened 2016-02-05
Improve routing click handler logic

dak:
Prevent the page from incorrectly refreshing if the target is not an anchor tag (if, for example, an anchor tag is the parent of an img tag).  We now traverse up the DOM more intelligently to find any relevant anchor tags.

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -6,18 +6,6 @@ import appView from '~/components/shell/shell';
 const EXTERNAL = /^((f|ht)tps?:)?\/\//;
 const MAILTO = /^mailto:(.+)/;
 
-function rootEl(path, Element) {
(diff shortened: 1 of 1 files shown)


src/app/main.js
```diff
@@ -61,6 +49,16 @@ class App {
             pushState: true
         });
 
+        function findAncestor(el, Element) {
```
RoyEJohnson: This might belong in the $ file. It's akin to jQuery's `closest()`

dak: Hmm, maybe, although I'm not sure it would actually be used anywhere else (how often do you use `closest()`).

RoyEJohnson: Never, if I can help it. If we need such a thing, we can move it later.

PR #21, opened 2016-02-05
Implement Details and Subjects pages

RoyEJohnson:
Using Backbone model for fetched data and app state
Implemented node management to handle served/changeable data
Replaced $.toggleClass with built-in classList.toggle


src/app/helpers/backbone/node-manager.js
```diff
@@ -0,0 +1,116 @@
...
+        this.managedNodes = {};
+
+        this._addNode(rootNode);
+        for (let node of rootNode.querySelectorAll('[data-manager]')) {
+            this._addNode(node);
+        }
+
+        for (let key in this.managedNodes) {
+            if (!(key in this.bindings)) {
+                window.console.warn('No binding for:', key);
```
dak: I think it's best to avoid console statements in the code.  I'm assuming you made it `window.console.warn` instead of `console.warn` to get around ESLint throwing an error?

dak:
image

I don't think it's necessary/useful to include `without Knockout` in the commit message, since there's no code in master that uses knockout.

src/app/pages/subjects/category-section/book/book.js
```diff
@@ -0,0 +1,52 @@
...
+        this.templateHelpers = {
+            coverUrl: data.coverUrl
+        };
+        this.data = data;
+        this.model = model;
+        this.nodeManager = new NodeManager(model);
+        this.nodeManager.manage.css(
+            'selected-book',
+            ['selectedBook'],
+            (value) => ({selected: value === data})
```
dak: NodeManager seems like a LOT of code to just manage CSS class.

RoyEJohnson: NodeManager manages DOM nodes. I pre-defined some managers, but the intent is for it to be entirely customizable.
There's a common pattern of program state being reflected in the view: A DOM node's behavior is supposed to be controlled by fetched data (as opposed to static data). We need to keep track of when the node is re-rendered and when the data changes. This comes down to two variables: the node, the data item; plus a bit of code to plug the data item into to get all the data involved, and a boilerplate bit of code to act on that data.
In the HTML, you add a `data-manager=`*nodeName* attribute to the node to tie it to a manager.
In the JS code, you pass the node name, an array of model attribute names, and a function that takes the model attribute values (like require) and returns the final data to be used in the node. (The third argument defaults to the identity function, if you just want the attribute value itself)
To set HTML contents of a node, you'd pass them to `nodeManager.manage.html()`
In onRender, you typically call `nodeManager.updateNodes(this.el)` to refresh its ability to find nodes by name.
The node manager handles updating the node when the value changes, and the one call in onRender handles resetting state in the case of a re-render, but if your template doesn't contain information that varies over the life of the view (and it shouldn't) then you won't need to re-render anyway.

src/app/components/get-this-title/get-this-title.scss
```diff
@@ -0,0 +1,17 @@
...
+.get-this-title {
+    background-color: white;
+
+    .option {
+        color: color(cyan);
+        cursor: pointer;
+        width: 25rem;
+
+        .fa {
+            color: $darker-cyan;
```
dak: Try to use the global variables.  If you need to slightly darken a global, use SCSS's built-in `darken()` function.  For example: `color: darken(color(cyan), 5%);`

src/app/helpers/$.js
```diff
@@ -1,10 +1,15 @@
 let $ = {};
 
-$.toggleClass = function toggleClass(el, name) {
-    if (el.classList.contains(name)) {
-        el.classList.remove(name);
+$.setText = (node, value) => {
+    node.innerHTML = '';
+    node.appendChild(document.createTextNode(value));
```
dak: Why wouldn't you just do `node.textContent = value` and eschew this function entirely?  This seems to be doing a fair bit more than it implies as well, since it's effectively emptying the node first (`innerHTML = ''` isn't the best way to do that either).

src/app/pages/subjects/subjects.js
```diff
@@ -0,0 +1,167 @@
...
+                        coverUrl: book.coverUrl
+                    }))
+            }
+        ));
+
+    return {
+        filterButtons,
+        booksByCategory
+    };
+};
```
dak: Should all this dummy data be moved into a separate model that would also be used to load it?

src/app/pages/subjects/subjects.js
```diff
@@ -0,0 +1,167 @@
...
+        });
+
+        this.templateHelpers = {};
+        setTimeout(() => {
+            this.model.set('bookIndex', jsonData());
+        }, 800);
+
+        setTimeout(() => {
+            this.model.set(dummyData());
+        }, 300);
```
dak: I think simulating data loading on a delay is fine for testing, but I'm not sure it should find its way into master (unless it's part of an actual test).

src/app/pages/subjects/subjects.js
RoyEJohnson: I don't know. I hadn't planned to use a separate model to load it, I was expecting this model to do so.

dak: Since this is basically a Controller, it might be good to separate out the actual Model into a separate file to maintain a MVC-like structure.  It's possible (likely?) that this data will be relevant to other parts of the site as well.

RoyEJohnson: The "View" is a controller, but the Model is a Model, right? I have a Model into which the data gets loaded before I use it. The variable defining the data is just a stand-in and will go away when the data source is hooked up to populate the Model. Are we expecting other modules to do the data loading and to pass me a Model?

dak:
Merging, although I think there's some improvements that still need to be made on these pages.

From a design/usability perspective:
1) The styling doesn't seem to fully match the mockups, as the floating download nav when viewing a book covers up a lot of content (see the 2 screenshots below):
screen shot 2016-02-10 at 9 55 57 pm
screen shot 2016-02-10 at 9 56 06 pm

2) There's a few issues with accessibility, including properly marking the dropdown menus with aria tags, and including alt properties on all img tags.

3) I think there is supposed to be more margin/padding around pages.  For example, a book's title appears to sit too close to the nav menu (there also is no space before the footer):
screen shot 2016-02-10 at 10 02 21 pm

From a code perspective:
4) The `setHtmlToMember` function in subjects.js looks like it's fighting the templating engine.  You could probably simplify/remove that entirely.

5) There's a lot of custom color variables in certain files.  It might be best if we try to restrict ourselves to using the colors pre-defined globally, as that helps to change things and also to keep the brand and look-and-feel more consistent between pages.

6) A lot of the code creating models that is sitting in a "View" (or Controller or whatever it is) should probably be moved out into its own file under /models.  That will help keep a separation of concerns and allow more code reuse.

7) There appears to be a mixing of model and view/controller knowledge in some places.  For example, in subjects.js you have ` this.model = new BaseModel({viewIsReady: false});`.  This suggests that the model is storing information about the status of a view, which is breaking the separation of concerns.

8) In components/buttons.scss you've defined styling for `.cta > .btn`.  It's not clear what that is for or why a button would be in `.cta`.  Global components should be for defining a consistent style for common components reused throughout the app.  I'm not sure that's the case here.

src/app/pages/details/details.scss
```diff
@@ -0,0 +1,228 @@
...
+        .links > a {
+            display: block;
+            margin-top: 1rem;
+        }
+    }
+}
+
+a,
+.tab > label,
+.option {
```
dak: Eek, I missed this, but it's affecting EVERY anchor tag on the entire page!  This all needs to be moved inside a .details block.

src/app/pages/details/details.scss
```diff
@@ -0,0 +1,228 @@
...
+
+                &:last-child > label {
+                    border: 0;
+                }
+            }
+        }
+    }
+}
+
+.endorsement .endorsement {
```
dak: Why are there elements in an element that both have the same class name?

src/styles/components/buttons.scss
```diff
@@ -26,7 +26,16 @@
...
+.cta > .btn {
+    background-color: color(orange);
+    border: 0;
+    color: color(neutral-lightest);
+    font-variant: small-caps;
+    margin: 0 auto;
+    padding: 1.5rem 2rem;
+}
+
+@each $name, $type in $color-types {
```
dak: Hmm, this is what broke a lot of colors.  It looks like it was accidentally changed from `$os-colors` to `$color-types`.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
+<style scoped>
+    @import url(/app/pages/details/details.css);
+    @import url(https://fonts.googleapis.com/css?family=Lato:400,900);
+</style>
+<section class="floating-menu">
+    <div class="title">Get this title</div>
+    <div class="get-this-book"></div>
+    <div class="notes">
+        <sup>*</sup>Our web view is Section 508 Level AA accessible.
```
dak: I think this should be `web site` not `web view`.  Also, we're not AA accessible, we're only Level A accessible.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
...
+    </div>
+    <div class="cta">
+        <p class="interested-text">
+            Using this book, or interested in adopting it for your
+            course?
+        </p>
+        <a href="/adoptions?{{title}}" class="btn">adopt this book</a>
+        <a class="secondary btn">get interested</a>
```
dak: If these are intended to be all lowercase, it might be more appropriate to use CSS to `text-transform` it to all lowercase, and code it in the document using more normal English capitalization.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
...
+    </div>
+</section>
+<section class="block book-info">
+    <div class="info-row">
+        <img src="https://placeholdit.imgix.net/~text?txtsize=33&txt=Sub-ject&w=200&h=180" />
```
dak: Here's a good example of where a global model could be useful.  You are already defining the book's cover URL in subjects.js where you create the model.  This could use that model and then be changed to just `{{cover}}`.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
...
+        <section class="details">
+            <h1 class="title">{{title}}</h1>
+            <div class="tabs">
+                <a class="tab"><label>Table of Contents</label></a>
+                <a href="#resources" class="tab"><label>Resources</label></a>
+                <a href="#details" class="tab"><label>Details</label></a>
+                <a href="#errata" class="tab"><label>Errata</label></a>
```
dak: label tags are for form elements.  This isn't a form.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
...
+        <div class="blurb">
+            <img src="https://placeholdit.imgix.net/~text?txtsize=33&txt=Logo&w=200&h=45" />
+            <h4>Sapling Learning</h4>
+            <p class="text">
+                Bacon ipsum dolor amet incididunt aliqua salami tongue fugiat
+                anim. Andouille sunt labore sint, incididunt turkey tri-tip
+                swine pork chop jowl cow fatback spare ribs shank enim. Commodo
+                cow laborum ad adipisicing spare ribs excepteur.
+            </p>
+        </div>
```
dak: Here's another good place for a model rather than repeating these identical blobs of code numerous times.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
...
+        <p>
+            All OpenStax textbooks undergo a rigorous review process. However, like
+            any professional-grade textbook, errors sometimes occur. The good part
+            is, since our books are web-based, we can make updates periodically. If
+            you have a correction to suggest, submit it here. We'll review your
+            suggestion and make necessary changes.
+        </p>
+        <div class="links">
+            <a>Errata list &gt;</a>
+            <a>Suggest a correction &gt;</a>
```
dak: If the `>` symbols are for stylistic purposes only, it may be more appropriate to add with a CSS pseudo element.  Otherwise it's quite awkward for users of screen readers.

src/styles/variables/colors.scss
```diff
@@ -79,3 +79,12 @@ $link-color-dark: color(primary);
 // Background Colors
 $background-color-light: color(neutral-lightest);
 $background-color-dark: color(neutral-dark);
+
+// Other things
```
dak: `Other things` probably isn't the most helpful comment.

src/app/pages/details/details.js
```diff
@@ -0,0 +1,76 @@
...
+            title,
+            blurb: `College Physics meets standard scope and sequence requirements for a
+                two-semester introductory alegabra-based physics course. The text is
+                grounded aliqua salami tongue fugiat anim. Andouille sunt labore sint,
+                incididunt turkey tri-tip swine pork chop jowl cow fatback spare ribs shank
+                enim. Commodo cow laborum ad adipisicing spare ribs excepteur.`,
+            mainAuthorTitle: 'Senior Contributing Authors',
+            mainAuthorHtml: `
+                 <p>Dr. Paul Peter Urone, California State University Sacramento</p>
+                 <p>Dr. Roger Hinrichs, State University of New York, College at Oswego</p>
```
dak: Would it be better for the model to just have the raw data, and any styling/html can be done in the template displaying it?

src/app/pages/details/details.js
```diff
@@ -0,0 +1,76 @@
...
+                    iconUrl: 'http://wwwimages.adobe.com/content/dam/acom/en/legal/images/badges/Adobe_PDF_file_icon_32x32.png',
+                    downloadUrl: '/downloads'
+                }
+            ]
+        });
+    }
+
+    onRender() {
+        this.el.classList.add('text-content');
+        this.regions.getThisTitle.show(this.getThisTitle);
```
dak: If this view is being completely redrawn, any views within it probably should as well.  Is creating the `getThisTitle` view in this view's constructor an attempt to save state, or is it just to avoid re-creating it?  Note: it will still be re-rendered from `.show()`.

src/app/pages/details/details.hbs
```diff
@@ -0,0 +1,172 @@
+<style scoped>
+    @import url(/app/pages/details/details.css);
+    @import url(https://fonts.googleapis.com/css?family=Lato:400,900);
+</style>
+<section class="floating-menu">
```
dak: `section` tags are defined on MDN as the following:
> The HTML &lt;section&gt; element represents a generic section of a document, i.e., a thematic grouping of content, typically with a heading. Each &lt;section&gt; should be identified, typically by including a heading (&lt;h1&gt;-&lt;h6&gt; element) as a child of the &lt;section&gt; element.

There are a large number of section elements throughout this template that appear to be inappropriate.  They should probably just be `div`s.

src/app/pages/subjects/category-section/book/book.hbs
```diff
@@ -0,0 +1,7 @@
+<img src="{{coverUrl}}" />
```
dak: All `img` tags should have an `alt` property.

src/app/pages/subjects/category-section/book/book.js
```diff
@@ -0,0 +1,53 @@
...
+
+        this.model.on('change:selectedBook', () => this.setState());
+    }
+
+    setState() {
+        this.el.classList.toggle('selected', this.model.get('selectedBook') === this.data);
+    }
+
+    onRender() {
+        this.el.classList.add('cover');
```
dak: Maybe this should be added to the BaseView so it can be added as a property.  Currently all views that are injected are wrapped in a `div` since that's the easiest way to manage/append them.

I had been considering trying to create them using a documentFragment instead so that extra div wasn't created (which would break this); however, since every template should currently be wrapped in a namespace (since scoped styling and the Shadow DOM aren't fully supported yet), I could see this being cleaner than adding the namespaced div into the template itself.

Unfortunately, until scoped styling is widely supported, I think this may break on browsers that DO support scoped styling.

Using this convention, the view is being created with the following DOM structure:
```html
<div class="cover">
    <div>
        // content (including @import with scoped styles)
    </div>
</div>
```
Since scoped styling only applies to all content inside the parent element, which would be the inner `<div>`, there'd be no way to style the outer div with the `cover` class applied to it.

Therefore, I think for the time being it is better to wrap the template in `<div class="cover">` as is done on other pages.

src/app/pages/subjects/subjects.js
```diff
@@ -0,0 +1,149 @@
...
+
+            if (this.model.get('viewIsReady') &&
+                this.model.has('bookIndex')) {
+                setHtmlToMember('page-description', 'page_description');
+                setHtmlToMember('ds1-head', 'dev_standard_1_heading');
+                setHtmlToMember('ds1-body', 'dev_standard_1_description');
+                setHtmlToMember('ds2-head', 'dev_standard_2_heading');
+                setHtmlToMember('ds2-body', 'dev_standard_2_description');
+                setHtmlToMember('ds3-head', 'dev_standard_3_heading');
+                setHtmlToMember('ds3-body', 'dev_standard_3_description');
```
dak: This function seems to be fighting with the templating engine.  Why not let it handle updating DOM elements?

src/app/pages/subjects/subjects.js
```diff
@@ -0,0 +1,149 @@
...
+    regions: {
+        filterButtons: '.filter-buttons',
+        bookViewer: '.book-viewer'
+    }
+})
+export default class Subjects extends BaseView {
+
+    constructor() {
+        super();
+        this.model = new BaseModel({viewIsReady: false});
```
dak: The Model should have no knowledge of the state of the view.  The view's state should be stored by the view itself.  Otherwise it's breaking the separation of concerns.

src/app/pages/subjects/subjects.scss
```diff
@@ -0,0 +1,155 @@
...
+        text-align: center;
+    }
+
+    .subhead {
+        font-size: 1.5rem;
+    }
+
+    .feature-block {
+        display: flex;
+        font-size: initial;
```
dak: `initial` is setting the font-size to its default value, but I'm not seeing where it's being changed?  Does this line have any effect?

src/app/pages/subjects/subjects.scss
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    .book-viewer {
+        background-color: $light-gray;
+        padding: 5rem 10rem;
+
+        .subject {
+            color: $text-black2;
+            font-size: 120%;
```
dak: Should this be `rem`?

RoyEJohnson: Your comments are appreciated, and I'll put together a branch to fix the
CSS issues. But you're going to drive yourself crazy reviewing the HTML and
content. That's just scaffolding. I'm expecting the design guys to produce
the real HTML and styling. Maybe that's a misplaced expectation on my part?

On font size percents: Sometimes I just want to indicate that the font
should be a bit bigger than it otherwise would be. Do you consider that a
bad practice? I went through and changed most of them to equivalent rems
(the one you commented on I apparently overlooked), but in my mind, it's
actually correct to say that I just want them bigger. your view?


On Wed, Feb 10, 2016 at 10:43 PM, Derek Kent <notifications@github.com>
wrote:

> In src/app/pages/subjects/subjects.scss
> <https://github.com/openstax/os-webview/pull/21#discussion_r52564070>:
>
> > +                padding: 1rem 2rem;
> > +
> > +                &.selected {
> > +                    color: $text-black;
> > +                }
> > +            }
> > +        }
> > +    }
> > +
> > +    .book-viewer {
> > +        background-color: $light-gray;
> > +        padding: 5rem 10rem;
> > +
> > +        .subject {
> > +            color: $text-black2;
> > +            font-size: 120%;
>
> Should this be rem?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openstax/os-webview/pull/21/files#r52564070>.
>



-- 
Roy Johnson
Javascript Developer
OpenStaxx

src/app/pages/subjects/subjects.scss
RoyEJohnson: Yes, if you toggle it in the debugger, the font gets smaller. Off the top of my head, I don't know where it's being changed, either.

src/app/pages/details/details.scss
RoyEJohnson: I see you fixed this already. :+1:

src/app/pages/details/details.js
RoyEJohnson: In the design review, some things are plain text (like the author titles) and some things allow for HTML editing because the structure isn't well-defined in advance.

src/app/pages/details/details.js
RoyEJohnson: I did it that way because I thought create-once render-many. In fact, there should be no re-rendering, so it doesn't matter. I'll move it.

src/app/pages/details/details.hbs
RoyEJohnson: I did give the appropriateness of the section tag some thought as I applied it. Each section is separately thematic, although the endorsement isn't so much a grouping, so I've demoted it to div. All the others strike me as sections. The floating menu is a thematic grouping, with a .title that functions a heading, while all the others begin with an `h1`.

src/app/pages/subjects/category-section/book/book.js
RoyEJohnson: Point taken. For this particular example, there is no scoped stylesheet. (Scoping stylesheets to repeated elements performs badly.) As a rule, I will not apply a class to `this.el` that is used in the scoped sheet for the component.

src/app/pages/subjects/subjects.js
RoyEJohnson: Now that I'm actually fetching data, I've got a better notion of how to work with Backbone models. `viewIsReady` will go away with my subjects-fetch branch, which will follow the branch I've made to fix up all these issues.

PR #25, opened 2016-02-11
Complimentary copy

RoyEJohnson:
Minimal layout and functionality for the form

### DIFF
--- a/src/app/pages/comp-copy/comp-copy.hbs
+++ b/src/app/pages/comp-copy/comp-copy.hbs
@@ -1,0 +1,61 @@
+<style scoped>@import url(/app/pages/comp-copy/comp-copy.css);</style>
+<h1>Request a copy of "Principles of Economics"</h1>
+<div class="comp-copy">
+    <div class="book-blurb">
+        <img class="book-cover" />
+        <div class="book-description">
+        </div>
+    </div>
+    <div>
+        <form>
+            <h2>Comp Copy Request Form</h2>
+            <p>Which comp would you like?</p>
+            <fieldset class="labeled-buttons">
+                <label>
+                    <input checked="checked" name="book_format" type="radio" value="true" />
+                    Premium iBooks Textbook Comp
+                </label>
+                <label>
+                    <input name="book_format" type="radio" value="false" />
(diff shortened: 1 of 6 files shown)


dak:
It looks like the first two commits in this PR have already been merged to master and can be removed.

src/app/pages/comp-copy/comp-copy.hbs
```diff
@@ -0,0 +1,61 @@
...
+                    <span>Number of students this semester</span>
+                    <input name="number_students" type="number" required />
+                </label>
+            </fieldset>
+            <label>
+                <p>Other Committee Members</p>
+                <textarea cols="50" name="other_members" rows="4"></textarea>
+            </label>
+            <div>
+                <input type="submit" value="Submit" role="button" aria-disabled="false">
```
dak: `aria-disabled="false"` is the default value for `aria-disabled`, so it's not necessary to explicitly include that here unless the state is actively being toggled (and in that case, it should probably be set to true here, since presumably the form would not be submittable until data was input into required fields).

Source: https://www.w3.org/TR/wai-aria/states_and_properties#aria-disabled

src/app/pages/comp-copy/comp-copy.hbs
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+    <div>
+        <form>
+            <h2>Comp Copy Request Form</h2>
+            <p>Which comp would you like?</p>
+            <fieldset class="labeled-buttons">
+                <label>
+                    <input checked="checked" name="book_format" type="radio" value="true">
+                    Premium iBooks Textbook Comp
+                    </label>
```
dak: /label tag doesn't align with opening label tag.

src/app/pages/comp-copy/comp-copy.scss
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    > div {
+        flex-grow: 1;
+        margin: 1rem;
+        max-width: 50%;
+    }
+
+    form {
+        border: thin solid lightgray;
```
dak: Is there a global color value that can be used in place of `lightgray`?

PR #26, opened 2016-02-11
Fetch data for Subjects page

RoyEJohnson:
Updated subjects to fetch and display data from our server. CORS is an
issue (also, server may not have any data at a given time).
Plugged a little more dummy data into details page instead of flat HTML.
Also folded in fix-ups


dak:
It looks like the `Implement Subjects page` and `Regularize form names` commits found their way into this PR as well, although they have already been merged to master.  This looks like it's good to merge once those commits are removed.

PR #28, opened 2016-02-12
Implement model for fetching page info

RoyEJohnson:
Move the model that fetches page information into a separate module.
Pages can be fetched as needed, referenced by `meta.type`.

### DIFF
--- a/src/app/models/pagecollection.js
+++ b/src/app/models/pagecollection.js
@@ -1,0 +1,51 @@
+import BaseModel from '~/helpers/backbone/model';
+import BaseCollection from '~/helpers/backbone/collection';
+import {props} from '~/helpers/backbone/decorators';
+
+let pageUrl = 'https://oscms-dev.openstax.org/api/v1/pages';
+
+@props({
+    url: pageUrl
+})
+class PageModel extends BaseModel {}
+
+@props({
+    url: pageUrl
+})
+class PageCollection extends BaseCollection {
+
+    findMetaType(name) {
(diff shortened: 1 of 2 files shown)


RoyEJohnson:
This is a first cut at it. I'm just feeling my way around here, so I'll take suggestions. Issues I had to handle:  

1. finding a model in the collection by a sub-member of an attribute
2. waiting for the initial fetch to be complete

src/app/models/pagecollection.js
```diff
@@ -0,0 +1,39 @@
+import BaseModel from '~/helpers/backbone/model';
+import BaseCollection from '~/helpers/backbone/collection';
+
+let urlOption = {
+        url: 'https://oscms-dev.openstax.org/api/v1/pages'
+    },
+    PageModel = BaseModel.extend(urlOption),
```
dak: You can use the new ES2015 class syntax rather than `extend`ing the model.  For example:
```javascript
@props({
    url: 'https://oscms-dev.openstax.org/api/v1/pages'
})
class PageModel extends BaseModel {
    
    initialize() {
        // initialize code
    }

    otherMethods() {
        // code
    }

}
```

dak:
@mwharrison @rich-hart How will page IDs be handled? Will they be static (and ideally based on strings that are related to their path)? Or will they be dynamic numbers?

rich-hart:
Page ids for the most part will be static, but the system is not set up to guarantee this in all cases.  I’ll make a issue of this to strictly map pages to unique static ids.

PR #37, opened 2016-02-22
Improve Home Page Responsiveness and UX

dak:
- [x] Make bucket and quote views global components
 - [x] Add CSS "arrows" to bucket and quote views
 - [x] Improve bucket and quote view's responsiveness for all resolutions
 - [x] Improve education banner view's responsiveness for all resolutions
 - [x] Update global color scheme
 - [x] Resolve issue that caused event delegation in views to sometimes fail to work
 - [x] Improve accessibility and ensure WCAG2A compliance
 - [x] Add US History banner video
 - [x] Disable video playback on mobile devices and under 780px width
 - [x] Complete tablet styling for all subject banners
 - [x] Fix incorrect layouts in Firefox due to scoped styling
 - [x] Disable loading of all banners except US History (will enable other banners as videos are added)


src/app/components/quotes/quotes.hbs
```diff
@@ -1,36 +1,36 @@
 <style scoped>
-    @import url(/app/pages/home/quotes/quotes.css);
+    @import url(/app/components/quotes/quotes.css);
 </style>
 <div class="quotes">
     <div class="quote-bucket right">
-        <img src="/images/home/quotes/juan.jpg" width="175" height="230" />
+        <div class="image juan"></div>
```
RoyEJohnson: This is clearly placeholder text/images. Do you want "juan" class, or something more structural like first-quote? Or maybe use pseudo-selectors like :nth-child on the quote-bucket div?

src/app/components/shell/header/header.hbs
```diff
@@ -1,56 +1,78 @@
...
 
-          <button class="expand-nav" type="button" role="button" aria-label="Toggle Navigation">
-            <span></span>
-          </button>
-        </div>
-      </nav>
+        <ul class="nav-menu">
+            <li class="nav-menu-item dropdown" role="presentation">
+                <a href="/subjects" role="menuitem" aria-haspopup="true">
+                    Subjects <span class="carrot"></span>
```
RoyEJohnson: meant to be class "caret"?

src/app/components/quotes/quotes.hbs
dak: I've been contemplating how to do this, since it's going to come from the server.  Ideally we'd be able to use `data-image="/images/home/quotes/juan.jpg"`, but no browser currently supports the CSS `attr()` function (which would allow us to do: `background-image: attr(data-image url)`).

So I think in the meantime our only option will be to define classes in CSS for each image in the quotes SCSS file, and then have the template add the class dynamically (probably by just basing it on the file name).

We could obviously just inject the image src if it wasn't a background-image, but, for design reasons, it needs to be.

src/app/components/shell/header/header.scss
```diff
@@ -74,606 +43,394 @@
...
-                    }
-                }
-            }
-        } //end expand-nav button
-    } //main-nav
+        li {
+            margin-right: 12rem;
+        }
+    }
 }
```
zroehr: Im not sure why you are rewriting this scss for the header? it seems what you are rewriting has nothing to do with the home page functionality. This is all relative to the header alone and I am responsible for this code.

src/app/components/shell/header/header.hbs
```diff
@@ -1,56 +1,78 @@
...
+                        <li><a href="/contact" role="menuitem">Contact</a></li>
+                        <li><a href="/give" role="menuitem">Give</a></li>
+                        <li><a href="/login" role="menuitem">Login</a></li>
+                    </ul>
+                </span>
+                <span class="cls"></span>
+            </li>
+        </ul>
+    </nav>
 </header>
```
zroehr: Im not sure why you are rewriting this html for the header? it seems what you are rewriting has nothing to do with the home page functionality. This is all relative to the header alone and I am responsible for this code. Please revert.

src/app/helpers/backbone/view.js
```diff
@@ -80,6 +79,70 @@ class BaseView extends Backbone.View {
         this.regions = new Regions(this.regions, this);
     }
 
+    delegate(eventName, selector, listener) {
```
RoyEJohnson: What's this about? I don't see that it's used, and it's not quite event delegation.

src/app/components/quotes/quotes.hbs
RoyEJohnson: I'm just talking about the name of the class for the div. Having it "juan" ties it to specific content, when it should be about function or structure. Not a real stopper, just something to consider.

dak: Yeah, I think each class will have to be specifically tied to each image.  So we'll need a:
```
.name {
    background-image: url(src);
```
for every single image to make it something that can be updated from the backend.  Other possible options include putting the style element directly on the div, but that has its own problems since it will likely mean mixing media-queries into the HTML too, which just starts getting really messy really fast.

I definitely don't like it, but I don't think we have another option for the time being.  If I'm missing a solution, please let me know.

src/app/helpers/backbone/view.js
dak: It's overriding a few Backbone methods (that are also used by Backbone.NativeView).  Basically I think the event delegation is slightly broken (which causes events to fail to attach correctly in some instances... you may not have run into it yet).

I'm planning on opening a PR for Backbone.NativeView as well to see if I can get this merged upstream, but that might take some time.

src/app/components/quotes/quotes.hbs
RoyEJohnson: It may get clearer when you're actually getting info from the back end. I think the server ought to deliver something named more like "first-image" than "juan". And your stuff would be similar. I'm done reviewing.

zroehr:
the mobile css does not follow @melaridi design. please see below.
PR code: 
screenshot 2016-02-24 13 40 37
screenshot 2016-02-24 13 41 48

@melaridi design:
screenshot 2016-02-24 13 42 54
screenshot 2016-02-24 13 43 40

Also missing flyout mobile submenu:
screenshot 2016-02-24 13 43 31

zroehr:
Again, all of this missing functionality was completed last week and resides in my branch

RoyEJohnson:
@zroehr Are you saying that master has correct behavior merged from your branch and this one doesn't, or that this one should include your branch's changes?

edwoodward:
Isn't this what Fred changed yesterday?

zroehr:
@RoyEJohnson master currently has correct behavior and this one doesnt.

edwoodward:
I think my comment is wrong. Fred's changes do not change the behavior 
of the Hamburger menu.

zroehr:
@edwoodward fred's changes yesterday related to the desktop secondary nav, the one with the rice logo. It seems those updates have made it in this branch. The missing functionality that I am referring to here is for the mobile header and nav overlay.

src/app/pages/home/education/education.scss
```diff
@@ -131,6 +195,13 @@
             text-transform: none;
             width: 27.76vw;
 
+            @include collapsed {
```
melaridi: `height: 7.88vw` on the `.btn` needs to go away. The element gets chopped on mobile 
screenshot 2016-02-24 15 44 48

src/app/components/buckets/buckets.scss
```diff
@@ -0,0 +1,184 @@
...
+                        $direction: left,
+                        $position: top 27% left -2.4rem,
+                        $color: color(green),
+                        $size: 2.4rem
+                    );
+
+                    @include collapsed {
+                        @include triangle(
+                            $direction: top,
+                            $position: top -2.4rem left 17%,
```
melaridi: Per design, should be `right: 5rem` on mobile/collapsed
For refernene: homepage PDF Design file: https://drive.google.com/open?id=0BxscqVIq0m0tZjk0d1FZUGplaG8

src/app/components/buckets/buckets.scss
```diff
@@ -0,0 +1,184 @@
...
+
+            .quote {
+                align-items: center;
+                display: flex;
+                flex-direction: column;
+                justify-content: center;
+
+                .blurb {
+                    margin: 0 auto;
+                    max-width: 50%;
```
melaridi: Needs to be  `max-width: 50%` on mobile

src/app/components/buckets/buckets.scss
```diff
@@ -0,0 +1,184 @@
...
+
+        &.allies {
+            background-color: color(orange);
+            background-image: url('/images/home/buckets/allies.jpg');
+            background-repeat: no-repeat;
+            background-size: cover;
+        }
+
+        &.full {
+            text-align: center;
```
melaridi: `text-align: left;` on mobile

src/styles/variables/responsive.scss
```diff
@@ -1,12 +1,8 @@
 $breakpoints: (
     phone: 480px,
-    tablet: 768px,
-    desktop: 1280px
+    tablet: 768px
 );
 
-// boxed view 1200px
-$boxed-width: 120rem;
```
melaridi: I think `boxed-width` should stay since it's being used throughout the site per Fred's designs.

PR #41, opened 2016-02-23
Get book information from server for catalog/subjects page

RoyEJohnson:


RoyEJohnson:
Michael is making cover images available, so I'll update the code in a bit.

RoyEJohnson:
I can pick up cover images later. This is a good spot to check in the page. There will be more new data later when I'm doing more with the details page.

PR #47, opened 2016-02-24
Make Subject selection work with router

RoyEJohnson:
Router only loads Subjects page if it is not the currently-loaded page.
Subjects page listens to the router and updates selected filter
accordingly.


RoyEJohnson:
Again, Maher could not make it merge. Maybe a permissions issue?

PR #49, opened 2016-02-26
Fix Adoption form issue that is breaking navigation

RoyEJohnson:
Main issue was leftover code from an iframe that I took out.
Some weirdness with regions remains, but isn't raising errors now.


RoyEJohnson:
Maher's merge permissions still need fixed

PR #50, opened 2016-02-26
Add endorsement quote and author lists to details page

RoyEJohnson:
Also correct book cover


RoyEJohnson:
Merged after Maher's ok.

PR #55, opened 2016-02-29
fix mobile header css & js

zroehr:
Items left: (will complete in smaller PR)
- [ ] add aria popup expanded values
- [ ] vertical align mobile menu
- [ ] make .caret an <a> link
- [ ] dropdown font size too big on mobile landscape
- [ ] double check firefox functionality

### DIFF
--- a/src/app/components/shell/header/header.js
+++ b/src/app/components/shell/header/header.js
@@ -273,51 +273,16 @@ class Header extends BaseView {
         }
     }
 
-    throttle(func, ms) {
-        var timeout, last = 0;
-
-        return function () {
-            var a = arguments,
-                t = this,
-                now = +(new Date()),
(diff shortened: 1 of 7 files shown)


src/app/components/shell/header/header.js
```diff
@@ -119,25 +119,62 @@ class Header extends BaseView {
...
             button.classList.toggle('expanded');
             button.setAttribute('aria-expanded', !!button.classList.contains('expanded'));
         });
     }
 
+    removeClass(array, className) {
+        let i = 0,
+            len = array.length;
+
+        for (; i < len; i++) {
```
RoyEJohnson: Note: you can use the `for (let item of array)` style for loop for cleaner loops.
Not a blocker, just information.

PR #65, opened 2016-03-03
Insert download links and resources

RoyEJohnson:
Get This Title links: View online, PDF, iBook hooked up
Details page: Instructor and Student Resources
Details page URL uses 'slug' instead of id, for more clarity


src/app/pages/details/details.hbs
```diff
@@ -93,16 +87,7 @@
                 {{/each}}
```
melaridi: I noticed you removed this `{{#each bookInfo.instructorResources}}`. Is this `{{/each}}` still needed?

PR #93, opened 2016-03-12
Add arrow up on open for dropdown

melaridi:
For both single and Multi select form dropdowns. 
Removed `h2` tags from adoption form. The form is self explanatory...


src/app/components/single-select/single-select.js
```diff
@@ -8,7 +8,8 @@ import {template} from './single-select.hbs';
 @props({
     template,
     regions: {
-        optionList: '.option-list'
+        optionList: '.option-list',
+        hasDropdown: '.has-dropdown'
```
RoyEJohnson: Are you doing anything with this region? A region is typically for putting sub-views into.

melaridi: I added class `open` on click

PR #118, opened 2016-03-17
Replace registered-trademark symbol with superscripted version

RoyEJohnson:
Done in boilerplate and in select options


RoyEJohnson:
Ugh. There's a styling issue on the filter buttons now.

PR #124, opened 2016-03-18
Add Higher Ed Page

dak:
- [x] Update Dependencies (fixes issue with ESLint and Babel)
 - [x] Make Product Boxes a global component
 - [x] Add Higher Ed Page


src/app/pages/higher-ed/higher-ed.scss
```diff
@@ -0,0 +1,134 @@
...
+        color: white;
+        padding: 3.57vw 8.21vw;
+
+        > .header {
+            font-size: 3.4vw;
+            font-weight: 900;
+        }
+
+        > .steps {
+            align-items: stetch;
```
RoyEJohnson: stretch? Maybe leave out if this typo isn't causing misbehavior.

zroehr:
@dak could you please review the higher ed page and double check that the 1200px grid is being maintained throughout the page. All content should fall within this max-width grid. Here is a link to the visual design: http://aizpfs.axshare.com/#p=higher_education_-_visual_design_-_guides  I also noticed the font size scales down when you resize the window. This should be removed and all text should stay same size throughout.

PR #140, opened 2016-03-24
Set CORS Headers for Dev

dak:


edwoodward:
Why is this needed since we have moved both front and back to 
https://oscms-dev.openstax.org/? Will this fix the current localhost not 
working issue?

dak:
@edwoodward Yes, this will fix the localhost issue.

PR #151, opened 2016-03-25
add link to source code

philschatz:
image

Adds a link to the OpenStax source code in the footer.

@openstax/ux: thoughts on placement/phrasing? I just put it somewhere but it could also be a circle, like the linkedin link but with a black-and-white :octocat:

### DIFF
--- a/src/app/components/shell/footer/footer.hbs
+++ b/src/app/components/shell/footer/footer.hbs
@@ -18,6 +18,9 @@
                     <a href="/accessibility-statement">Accessibility Statement</a>
                 </li>
                 <li>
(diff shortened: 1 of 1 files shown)


Fredasaurus:
I have no real preference here, though maybe it should go before contact.

rvnewell:
I would definitely prefer for it to stay in its current group (rather than grouped with social media).  Also agree with Fred that it’s probably better to put it before Contact (but no biggie.)

-Robert

philschatz:
Is linkedin more social than "Social Coding" :tm: (GitHub's tagline)? :smile: Rebased to have "Contact" be the last item.

PR #159, opened 2016-03-29
Add logged in user dropdown to header

zroehr:
- [x] add user login dropdown
- [x] change margin to padding on desktop dropdown
- [x] add hover state to items with dropdown on desktop header
- [x] add eventlistener to open login link in same window when not logged in

### DIFF
--- a/src/app/components/shell/header/header.js
+++ b/src/app/components/shell/header/header.js
@@ -211,6 +211,7 @@ class Header extends BaseView {
             header.classList.add('open');
             parentItem.classList.add('open');
             dropDownMenu.classList.add('open');
+            this.openThisDropdown(e);
         } else if (!e.target.classList.contains('back')) {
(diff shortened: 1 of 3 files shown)


src/app/components/shell/header/header.js
```diff
@@ -338,7 +339,10 @@ class Header extends BaseView {
             let loginItem = this.el.querySelector('.meta-nav .container .login a'),
                 userInfo = data[0];
 
-            if (userInfo && userInfo.username) {
+            if (!userInfo || userInfo.username === '') {
+                this.meta.loggedIn = false;
```
RoyEJohnson: This seems like it shouldn't work right, but maybe this is something I don't understand properly. The templateHelpers are used when the template is rendered, and then after rendering you're (possibly) changing the values used in templateHelpers. Is the templating system smart enough to re-render when its helper values change?

PR #160, opened 2016-03-30
fixed typo and removed title under image

edwoodward:


edwoodward:
@RoyEJohnson Please review. I'm pretty sure I removed the title under the image correctly, but I do not have a local instance to test with.

PR #164, opened 2016-03-30
Make menus for last row of books position static

RoyEJohnson:
That forces the window to have room for them.


zroehr:
Im going to address these issues in a separate PR. There have been page updates that will fix these issues in another way.

PR #180, opened 2016-04-04
Add preloader to website

zroehr:
Add preloader to website to prevent flash of unstyled content while dependencies are being loaded


src/app/components/shell/shell.js
```diff
@@ -9,22 +10,34 @@ import zendesk from '~/helpers/zendesk';
...
         this.render();
     }
 
+    preLoader() {
+        setTimeout(() => {
+            document.body.classList.remove('no-scroll');
+        }, 1000);
+        setTimeout(() => {
+            document.body.classList.add('loaded');
+        }, 2000);
```
dak: Is this just a temporary hack?  A 2 second delay on loading is probably enough to make it look like the site has loaded on machines near the server and with good internet connections, but it's not an accurate representation of the loading state.

PR #185, opened 2016-04-04
Fix Adoption, Details, and Subjects pages

RoyEJohnson:


src/app/pages/details/details.scss
```diff
@@ -74,6 +74,11 @@
         }
     }
 
+    // scss-lint:disable IdSelector
+    #student-resources-section {
+        padding-top: 5rem;
+    }
```
zroehr: Could we either add padding to the ally's section above this section or remove this so the spacing between the sections stay even? it seems that the scrolling offset is fine without the padding so im not sure this is needed.

RoyEJohnson: I'll take another look at it when I get in.

PR #191, opened 2016-04-05
Allies updates

RoyEJohnson:


src/styles/components/headers.scss
```diff
@@ -43,13 +43,12 @@ h4 {
     font-weight: 300;
     letter-spacing: 0.11rem;
     line-height: 3rem;
-    padding: 4rem 8rem;
+    padding: 2rem 2rem 11rem;
```
zroehr: Do hero's need this much padding on the bottom and so little on top? Also, I think the padding for the gutters on each side are 1.5rem (at least this is the value ive been using everywhere)

RoyEJohnson: The task was to "reduce white space in banner". (Trello card is https://trello.com/c/GZwwy4EW)
I eyeballed the spacing from the visual design doc (http://aizpfs.axshare.com/#p=allies_-_visual_design)
The particular numbers aren't sacred to me; have a look and see if you think they should be changed.

PR #211, opened 2016-04-11
Copy polyfills file along with system.js

RoyEJohnson:


dak:
This fixes IE11 loading.

gulp/tasks/copy.js
```diff
@@ -10,9 +10,12 @@ function copy() {
         systemjs = 'jspm_packages/system.js*';
     }
 
+    polyfilljs = 'jspm_packages/system-polyfills.js';
```
dak: `polyfilljs` is undefined, which is causing travis to error. It should probably be added to line 5 as well.

RoyEJohnson: Done.

PR #228, opened 2016-04-14
Make page loader recognize URL with #

RoyEJohnson:
This allows cross-page #links to work.

Because of dynamic loading, the location of a #link will likely change
from where it is when the page initially loads. To compensate, I’ve put
an interval that will re-adjust the scroll position a few times. A
fully correct solution would require each page to signal when it was
stable, which would require significant work on each page.

### DIFF
--- a/src/app/components/shell/shell.js
+++ b/src/app/components/shell/shell.js
@@ -1,4 +1,5 @@
 import BaseView from '~/helpers/backbone/view';
+import $ from '~/helpers/$';
 import {props} from '~/helpers/backbone/decorators';
 import header from './header/header';
 import footer from './footer/footer';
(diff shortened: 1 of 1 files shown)


src/app/components/shell/shell.js
```diff
@@ -37,7 +38,25 @@ class AppView extends BaseView {
...
+            if (hash) {
+                let attempts = 0,
+                    i = setInterval(() => {
+                        let destination = document.getElementById(hash.substr(1));
+
+                        if (destination) {
+                            $.scrollTo(destination);
+                        }
+                        ++attempts;
+                        if (attempts > 6) {
```
dak: Where does 6 come from?

RoyEJohnson: It's the arbitrarily-chosen "few times" I mention in my PR comment.

PR #242, opened 2016-04-18
Scroll form dropdowns

RoyEJohnson:
Created a touch-scroller object
Used it in bios and proxy select widgets
Created a browser test so redrawing can be done only on IE11 where it
is needed
Put in a noticeable delay.

### DIFF
--- a/src/app/helpers/$.js
+++ b/src/app/helpers/$.js
@@ -8,6 +8,8 @@ $.setAttr = (node, name, value) => {
     }
 };
 
+$.isIE11 = () => !(window.ActiveXObject) && 'ActiveXObject' in window;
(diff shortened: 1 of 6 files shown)


src/app/helpers/$.js
```diff
@@ -8,6 +8,8 @@ $.setAttr = (node, name, value) => {
     }
 };
 
+$.isIE11 = () => !(window.ActiveXObject) && 'ActiveXObject' in window;
```
dak: :anguished::cold_sweat::sob:

RoyEJohnson: The many faces of Internet Explorer

PR #305, opened 2016-04-29
add home banner flyin/flyout animations back

zroehr:
Flyin/flyout banner animations were never meant to be removed. Reverting changes back.


Fredasaurus:
It's a peacock.

PR #317, opened 2016-05-02
Fix bugs on Impact, K-12, Foundation & Higher-Ed

melaridi:
- [x] Added `$boxed-width`  to Impact page logo's container 
- [x] Removed `vh` from all Pages with Hero. It caused the banner to jump in Android.
- [x] Fixed Foundations' page logo overlap in IOS.
- [x] Adjust H1 `line-height` in **hero.sccs**

### DIFF
--- a/src/app/pages/k-12/hero/hero.scss
+++ b/src/app/pages/k-12/hero/hero.scss
@@ -27,8 +27,8 @@ $hero-break-width: 880px;
         @include collapsed {
             align-items: stretch;
             background: none;
-            height: calc(100vh - 6rem);
-            max-height: 53rem;
+//            height: calc(100vh - 6rem);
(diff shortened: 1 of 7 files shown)


src/app/pages/k-12/hero/hero.scss
```diff
@@ -1,6 +1,11 @@
 @import 'variables';
 @import 'mixins';
 
+html,
+body {
+    height: 100%;
+}
```
dak: Is this necessary?  If so, it's a global change that will affect every page, so it should probably be moved to `main.scss` and verified that it doesn't break any other pages.

melaridi: my bad, I was testing and forgo to remove it.

PR #321, opened 2016-05-02
fix font rendering

zroehr:
fix font rendering on local devices & ie11 by self hosting fonts from fonts.com


dak:
@edwoodward Do we have a license to self-host these web fonts?

PR #346, opened 2016-05-09
add favicon

zroehr:
add favicon that tutor uses: https://github.com/openstax/tutor-server/blob/master/public/favicon.ico


dak:
`https://openstax.org/images/icons/favicon.ico` returns a 404.

PR #347, opened 2016-05-09
Use canonical names as values in Salesforce forms

RoyEJohnson:


src/app/pages/comp-copy/comp-copy.hbs
```diff
@@ -63,18 +63,18 @@
             <span>Book Requested</span>
             <select name="00NU00000053nzR" required />
                 <option value="" disabled selected>Please Select</option>
-                <option value="Anatomy and Physiology">Anatomy and Physiology</option>
+                <option value="Anatomy & Physiology">Anatomy and Physiology</option>
```
dak: Should this be `&amp;`?

dak:
It seems like it might be useful to make this into a model/data that can be loaded dynamically by the template, rather than repeating it in multiple files.

src/app/pages/comp-copy/comp-copy.hbs
RoyEJohnson: It took me a while to find an answer, but it looks like yes.

RoyEJohnson:
I actually have a file with the book data. The challenge is that not all the forms use all the titles.

PR #373, opened 2016-05-23
Fix margin on Get The Application button

RoyEJohnson:
Adoption status changes were superseded by another PR, so I have removed that commit here.


RoyEJohnson:
Added commit to fix sorting on Partners page (per request from Courtney)

PR #379, opened 2016-05-26
add chemistry banner

zroehr:
add responsive css & js for chemistry banner


zroehr:
waiting on copy update from @CourtneyRay

PR #386, opened 2016-06-03
simplify install by not relying on globals

philschatz:
the steps to install and get up and running should now be:

1. install https://github.com/creationix/nvm
2. `nvm use`
3. `npm install`
4. `npm start`

And then, if we change things, no need to update the deployment scripts if we change versions of things (maybe add a `npm run build` script?)

/cc @cnuber


package.json
```diff
@@ -34,6 +35,7 @@
     "gulp": "github:gulpjs/gulp#4.0",
     "gulp-autoprefixer": "3.1.0",
     "gulp-babel": "6.1.2",
+    "gulp-cli": "gulpjs/gulp-cli#4.0",
```
dak: I think this needs to be changed to just `"gulp-cli": "1.2.1",`

package.json
```diff
@@ -9,6 +9,7 @@
     }
   ],
   "scripts": {
+    "start": "gulp dev",
```
dak: This does not appear to work when `gulp-cli` is not installed globally or `gulp` is otherwise added as a PATH environment var or linked to

Making `gulp-cli` a dev dependency doesn't seem to make any difference either.

philschatz: The error I noticed (maybe it's not the same as the one you saw) was:

```
os-webview/gulp/tasks/watch.js:32
    gulp.parallel(...watchTasks)
                  ^^^

SyntaxError: Unexpected token ...
    at exports.runInThisContext (vm.js:53:16)
```

But running `nvm use` fixed it (new version of JS that supports `...`)

philschatz:
Rebased master into here 😄

package.json
dak: You're right that's an error with the version of nodejs.

The error I was seeing was that the `gulp` command simply can't be found when I try to set it up on a fresh system. It does seem like an older version of gulp or gulp-cli may have added a wrapper to `/usr/local/bin/gulp` that makes everything work if you don't do the install on a fresh system.

PR #391, opened 2016-06-08
add news page structure

zroehr:


src/app/pages/article/article.js
```diff
@@ -0,0 +1,90 @@
...
+            let newTags = [];
+
+            for (let tag of data.tags) {
+                let spanObj = `<span>${tag}</span>`;
+
+                newTags += spanObj;
+            }
+
+            this.el.querySelector('.tags').innerHTML = newTags;
+        });
```
dak: It would probably be better to let the template handle all this rendering

dak:
@zroehr Merging #397 and then rebasing against master will fix the Travis error you're seeing (and all future builds will see).

src/app/pages/article/article.js
```diff
@@ -0,0 +1,91 @@
...
+
+                    newTags += spanObj;
+                }
+
+                this.el.querySelector('.tags').innerHTML = newTags;
+
+                /* disabling the following eslint requirements due to disqus_config requirements */
+                /* eslint no-unused-vars: 0 */
+                /* eslint camelcase: 0 */
+                let disqus_config = () => {
```
RoyEJohnson: Is this something that will be used in the future when we hook up Disqus for comments?

src/app/pages/article/article.js
```diff
@@ -0,0 +1,91 @@
...
+
+        this.regions.articles.append(new Articles(slug));
+
+        let d = document,
+            s = d.createElement('script');
+
+        s.src = '//openstax.disqus.com/embed.js';
+        s.setAttribute('data-timestamp', +new Date()); (d.head || d.body).appendChild(s);
+
+        let d2 = document,
```
RoyEJohnson: Just use `d` again rather than declaring a new copy.

src/app/pages/blog/articles/articles.scss
```diff
@@ -0,0 +1,91 @@
...
+            padding: 0;
+            flex-flow: row wrap;
+        }
+    }
+
+    .container {
+        display: flex;
+        flex-flow: row wrap;
+
+        @include desktop {
```
RoyEJohnson: This changes nothing, right?

src/app/pages/article/article.js
zroehr: disqus is already used for comments. per their documentation, this is required. it must be used in their embed.js script that is appended.

RoyEJohnson: Unless there's some really dark magic going on, I don't think their script will be able to see this variable, because it's local to the scope that closes immediately after you define it. Are you able to confirm that Disqus is using the values you set here?

src/app/pages/article/article.js
zroehr: when using `d` again, disqus fails to load. im not sure why.

PR #395, opened 2016-06-13
Add partners to Salesforce dropdowns

RoyEJohnson:
Add eMath and panOpen to dropdowns.


dak:
This definitely looks like something that would benefit from having this list in a single model that other views can use.

PR #398, opened 2016-06-14
Add recordo.js

philschatz:
Using [recordo.js](https://npmjs.com/package/recordo) this allows adding `?collect=true` to the URL which will log:

- browser/OS information
- Ajax requests
- clicks
- window re-sizes

1. Should it only be enabled for development builds?
2. It currently uses a hacky way of initializing because I don't know the correct place in the code to start it up; is there a better way to start it up?


dak:
This PR builds off #386, which needs to be merged before this can be.

PR #421, opened 2016-06-29
Fix Contact Us Comments box

RoyEJohnson:
Allow word wrapping and scroll bar for texture
Also make it use our default font and not so bold in the fields.


RoyEJohnson:
I've rebased this to change-ibook-to-download-on-ibooks, so that needs to be merged first.

PR #422, opened 2016-06-30
Refactor homepage banners

RoyEJohnson:
Use one template for all banners
Put common CSS into home.css so individual banner css files are less
busy.


RoyEJohnson:
Included in refactor

Issue #392, opened 2016-06-10
Why does openstax depend on files on rawgit.com?


philschatz:
There is a script tag in `index.html` that fetches `classList.js`

https://github.com/rgrove/rawgit/wiki/Frequently-Asked-Questions#i-need-guaranteed-100-uptime-should-i-use-cdnrawgitcom

PR #423, opened 2016-07-05
Refactor for version 2.0

dak:
Fixes: #392 

NOTE: ~18 issues remain impacting performance, maintainability, and/or creating a memory leak (run `gulp eslint` to see most of the remaining issues that have been identified).

Pages Confirmed Updated:
 - [x] 404
 - [x] About
 - [x] Accessibility Statement
 - [x] Adopters
 - [x] Adoption
 - [x] ~~Adoption Confirmation~~
 - [x] Article
 - [x] Blog
 - [x] ~~CMS~~
 - [x] Comp Copy
 - [x] ~~Comp Copy Confirmation~~
 - [x] Confirmation
 - [x] Contact
 - [x] ~~Contact Thank You~~
 - [x] Details
 - [x] Faculty Confirmation
 - [x] Faculty Verification (Duplicate of Finished Verify)
 - [x] Finish Profile
 - [x] ~~Finished No Verify (Near Duplicate of Finished Verify... combine into one)~~
 - [x] ~~Finished Verify (Duplicate of Faculty Verification)~~
 - [x] Foundation
 - [x] Higher Ed
 - [x] Home
 - [x] Impact
 - [x] Interest
 - [x] ~~Interest Confirmation (Near Duplicate of other Confirmation pages)~~
 - [x] K-12
 - [x] License
 - [x] ~~Mass Renewal~~
 - [x] Partners
 - [x] Renew
 - [x] ~~Renewal~~
 - [x] Subjects
 - [x] Support
 - [x] Terms of Use

Components to Refactor

 - [x] Buckets
 - [x] Quotes
 - [x] ~~Single Select~~
 - [x] ~~Multi Select~~
 - [x] Select
 - [ ] Header
 
Additional:
 - [x] Most confirmation pages probably need to be combined into a single confirmation page, they're also duplicating a lot of CSS (they even use the same class, but load multiple files)
 - [x] Add testing framework (this might be expanded further with something like Nightmare)
 - [ ] Home Page performance


RoyEJohnson:
@dak I'm likely going to need to have you walk me through a fix or two so that I understand how you expect things to be done. It looks like you're suggesting that every state change result in a render; if that's the paradigm you like, it makes sense to look into a modern DOM-differencing framework like https://github.com/yoshuawuyts/choo

src/app/pages/partners/partners.html
```diff
@@ -7,7 +7,11 @@
     </div>
     <div class="icons">
         <img class="strips" src="/images/components/strips.svg" height="10" alt="" role="presentation">
-        <div id="filter" class="filter" skip="true"></div>
+        <div id="filter" class="filter">
+            <div class="filter-buttons">
+                <div each="button in model.filterButtons" class="filter-button{button === history.state.filter ? ' selected' : ''}">{button}</div>
```
RoyEJohnson: The AP button needs to have its superscripted registered trademark symbol.

src/app/components/select/select.js
```diff
@@ -5,6 +5,8 @@ import {description as template} from './select.html';
 const CONVERT_OPTIONS = Symbol();
 const CLOSE_DROPDOWNS = Symbol();
 const CONTROLLERS = Symbol();
+const STOP_SCROLLING = Symbol();
```
RoyEJohnson: This use of symbols to create static methods/members seems like it isn't buying you anything over just creating functions or variables outside the class (but inside the module). They're just private functions and variables.

dak: Correct. And they actually aren't truly private (it's possible to access them with `Object.getOwnPropertySymbols`). I try to put it on components that may be inherited/overloaded though to avoid name collisions and make it clear what should be considered an internal method without the `_` convention.

PR #438, opened 2016-07-19
Send email with Finish Profile form

RoyEJohnson:
Email is stored with cookies and will automatically be populated in a
hidden field.


src/app/pages/finish-profile/finish-profile.hbs
```diff
@@ -18,6 +18,7 @@
                 <input type=hidden name="first_name" />
                 <input type=hidden name="last_name" />
                 <input type=hidden name="user_id" />
+                <input type=hidden name="email" />
```
dak: Where is the value for this coming from? It looks to me like it's just always a valueless hidden input.

RoyEJohnson:
Included in refactor

PR #439, opened 2016-07-20
Fix TOC numbering on books with units

RoyEJohnson:
Also: remove links to removed regions on Details (mostly affects Coming
Soon books, which have no Resources sections)
Also: On Subjects, update Location when filter is selected


RoyEJohnson:
Included in refactor

PR #447, opened 2016-08-05
Use CMS data for Higher Ed and Home pages

RoyEJohnson:


src/app/pages/about/about.js
```diff
@@ -64,9 +64,7 @@ export default class AboutUs extends CMSPageController {
         };
 
         this.model = {bios: []};
-        this.queryPage = {
-            type: 'pages.AboutUs'
-        };
+        this.id=90;
```
dak: Minor nitpick: there's a single space on both sides of binary operators everywhere else in the code.

RoyEJohnson: updated

PR #470, opened 2016-08-17
Assorted change requests

RoyEJohnson:
Copy changes on Interest form
Update email fields to be of type “email”
Copy changes on AP page

### DIFF
--- a/src/app/pages/adoption/adoption.html
+++ b/src/app/pages/adoption/adoption.html
@@ -47,7 +47,7 @@
                     </label>
                     <label>
                         <span>Email address</span>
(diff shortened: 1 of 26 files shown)


src/app/pages/adoption/adoption.html
```diff
@@ -47,7 +47,7 @@
                     </label>
                     <label>
                         <span>Email address</span>
-                        <input name="email" pattern="[^@]+@([^@\.]+\.)+[^@\.]+" type="text" required>
+                        <input name="email" pattern="[^@]+@([^@\.]+\.)+[^@\.]+" type="email" required>
```
dak: You actually don't need the pattern anymore. It's not supported with type `email` and no browser we support will be falling back to treating it as type `text`.

PR #476, opened 2016-08-18
Add Optimizely and Lucky Orange trackers to Release-2.0

RoyEJohnson:


dak:
I merged the hotfix version of this, but I'm curious to see if it even works (not to mention the script size of 1.2MB is justified).

I strongly suspect this isn't going to work in a single page app out of the box, and we'll have to do something similar to how GA was added in the analytics.js handler, where we explicitly handle navigating to a new page.

dak:
Maybe hold off on optimizely until we actually have something to test, since it sounds like that's going to take a fair bit of developer intervention to get to work:
https://help.optimizely.com/Build_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications#How_does_Optimizely_work_with_Single_Page_Applications.3F

It looks like LO can be included with GTM, so there should be no extra code necessary (it claims it can work with SPAs out of the box).

PR #478, opened 2016-08-22
Make give thank you page according to visual design

RoyEJohnson:
Make it a variant of the Give page (instead of the confirmation page)
since the design is almost the same.


RoyEJohnson:
Ugh. Not for master.

PR #623, opened 2016-11-01
Track file downloads

edwoodward:


RoyEJohnson:
[Location incorrect in Google Analytics if user uses back button](https://trello.com/c/mlRyoyJr/18-location-incorrect-in-google-analytics-if-user-uses-back-button)

PR #665, opened 2016-12-07
Verify apiOrigin is production before sending event to Google Analytics

edwoodward:


edwoodward:
[Filter QA and Dev out of Google Analytics](https://trello.com/c/yfbOCxMH/41-filter-qa-and-dev-out-of-google-analytics)

PR #668, opened 2016-12-09
Blocked all links under /l

edwoodward:


edwoodward:
[Block indexing of /l links via robots.txt](https://trello.com/c/m0canv7L/46-block-indexing-of-l-links-via-robots-txt)

PR #672, opened 2016-12-14
Changed Volume to Part

edwoodward:


edwoodward:
[College Physics ibooks links should say Part 1 and Part 2 instead of Volume 1 and Volume 2.](https://trello.com/c/e27E3AoR/64-college-physics-ibooks-links-should-say-part-1-and-part-2-instead-of-volume-1-and-volume-2)

PR #678, opened 2016-12-16
Handle capitalized book slugs

RoyEJohnson:
Force bookTitle (slug) to lowercase
route to 404 on error fetching book data

### DIFF
--- a/src/app/controllers/cms.js
+++ b/src/app/controllers/cms.js
@@ -22,7 +22,7 @@ class CMSPageController extends Controller {
 
                     this.onDataLoaded();
                 } catch (e) {
-                    console.log(e);
+                    this.onDataError();
                 }
             })();
         }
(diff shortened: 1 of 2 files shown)


RoyEJohnson:
[ensure links are never broken](https://trello.com/c/SVKyUHIb/76-ensure-links-are-never-broken)

philschatz:
Thanks for fixing the existing links and sending to a 404!

I can no longer comment on the card so I'll post here... This seems like it only fixes the proximate problem (existing links were broken in a certain way and that change was missed as part of QA) but does not address the title of the card ("ensure that links are never broken").

/cc @kajalp to see if this should be done by devs (as unit/integration tests) or by QA (as integration tests)

RoyEJohnson:
When we discussed the card in scrum, the consensus was that it would be better to handle it by hygiene - not changing slugs in the CMS - than to add the capability in the CMS to track multiple changes of slugs.
Outside of book details, we expect that some links will be broken, because we are removing some pages (AP, for example, and k-12 which is already gone).

edwoodward:
@philschatz Can we merge this?

philschatz:
Yeppers, I didn't mean to block the PR because of discussion about the Issue (I think I now have access to comment/discuss on the Trello card)

PR #689, opened 2017-01-04
Use Django api for login and logout

RoyEJohnson:


RoyEJohnson:
[BUG - New Accounts routed incorrectly.](https://trello.com/c/THBLa8VW/89-bug-new-accounts-routed-incorrectly)

PR #691, opened 2017-01-05
Fix profile and faculty access links

RoyEJohnson:


RoyEJohnson:
[ACCOUNTS - When you click on the hi, [name] you are taken to the old accounts profile page…the user needs to be delivered to the new accounts profile page.](https://trello.com/c/hLrHyLuq/91-accounts-when-you-click-on-the-hi-name-you-are-taken-to-the-old-accounts-profile-page-the-user-needs-to-be-delivered-to-the-new-)

PR #694, opened 2017-01-06
Donation to salesforce

edwoodward:


edwoodward:
[Donation page: Get email addresses from donation form and send them to Salesforce](https://trello.com/c/sSVxd22n/17-donation-page-get-email-addresses-from-donation-form-and-send-them-to-salesforce)

src/app/pages/give/give.js
```diff
@@ -81,6 +81,18 @@ export default class Give extends CMSPageController {
...
             this.regions.share.attach(new FbShare());
             this.model.isThanks = true;
+
+            this.el.querySelector('[name=first_name]').value = localStorage.getItem('donation:first_name');
+            this.el.querySelector('[name=last_name]').value = localStorage.getItem('donation:last_name');
+            this.el.querySelector('[name=email]').value = localStorage.getItem('donation:email');
+            this.el.querySelector('[name=phone]').value = localStorage.getItem('donation:phone');
+            this.el.querySelector('[name=Donation_Amount__c]').value = localStorage.getItem('donation:amount');
+            document.getElementById('donation-update').submit();
+            localStorage.setItem('donation:first_name', '');
```
RoyEJohnson: You should use `localStorage.removeItem` to remove them, rather than setting their values to empty strings.

src/app/pages/give/give.js
```diff
@@ -202,6 +214,12 @@ export default class Give extends CMSPageController {
         if (invalid) {
             event.preventDefault();
             this.update();
+        } else {
+            localStorage.setItem('donation:first_name', this.el.querySelector('[name=First_Name]').value);
```
RoyEJohnson: These names don't match the form elements, capitalization-wise, and you're using `AMT` for `Donation_Amount__c`

edwoodward: I used the names in the form where the values are entered for the local storage names. I added the donation: prefix so they did not clash with any other items in local storage. The salesforce form names are determined by Salesforce. Which would you like me to match?

src/app/pages/give/give.js
edwoodward: Good point. I will update.

src/app/pages/give/give.js
RoyEJohnson: It's the names in the querySelector need to be the same as the name attributes in the HTML form. You have
`<input type="hidden" name="first_name">`
but are querying
`this.el.querySelector('[name=First_Name]')`

edwoodward: First_Name is from the form at the top of give.html where the user enters the information. The input type name (first_name) is in the hidden form I am submitting to salesforce. I did not use the same names so the query would find the correct fields (First_Name vs first_name).

RoyEJohnson: Ok. That's my misunderstanding, then.

edwoodward:
@RoyEJohnson PR is updated

PR #699, opened 2017-01-10
Move login to sticky menu

RoyEJohnson:
Also hide sticky note if there is no content


RoyEJohnson:
[Add login option on sticky menubar](https://trello.com/c/xEmGf6DC/100-add-login-option-on-sticky-menubar)

PR #705, opened 2017-01-19
Fix capitalization in footer tagline

RoyEJohnson:


RoyEJohnson:
["Future" and "education" in tagline in footer should be lowercase.](https://trello.com/c/qDKZBJWa/48-future-and-education-in-tagline-in-footer-should-be-lowercase)

PR #706, opened 2017-01-19
Fix redirect parameter for login/logout links

RoyEJohnson:


RoyEJohnson:
[BUG - New Accounts routed incorrectly.](https://trello.com/c/THBLa8VW/89-bug-new-accounts-routed-incorrectly)

PR #717, opened 2017-02-01
Fix Donation page layout

RoyEJohnson:
2 columns of buttons on phone
2 rows of buttons on larger screens
IE layout matches other layouts
Fixed header text size


RoyEJohnson:
[[OSW2.design.05] Fix donation page design](https://trello.com/c/sbQfPMhp/13-osw2-design-05-fix-donation-page-design)

PR #733, opened 2017-02-09
centered header text

edwoodward:


edwoodward:
[Center header copy](https://trello.com/c/EokWnvg0/137-center-header-copy)

PR #737, opened 2017-02-09
Send BILL_* fields to Touchnet

RoyEJohnson:


RoyEJohnson:
[Send Bill_* fields from give form to populate in Touchnet](https://trello.com/c/Xd2CLPjo/138-send-bill-fields-from-give-form-to-populate-in-touchnet)

PR #741, opened 2017-02-13
Fix book price format

RoyEJohnson:


RoyEJohnson:
[Book price should display to the hundredths place on /subjects](https://trello.com/c/wcC0OiuF/148-book-price-should-display-to-the-hundredths-place-on-subjects)

PR #745, opened 2017-02-14
Fix TOC numbering

RoyEJohnson:
Number only if it has contents or has a numbered parent.


RoyEJohnson:
[Table of Contents is wrong for Alg & Trig.](https://trello.com/c/zTSmsvwv/146-table-of-contents-is-wrong-for-alg-trig)

PR #754, opened 2017-02-21
Improve errata form load time

RoyEJohnson:
Make supermodel cache results for 15 seconds
In analytics, use the book fetch to add partner information to the
lookup tables, reducing a second fetch of book data


RoyEJohnson:
@edwoodward Please look at the changes to analytics code closely. I'm confident that rolling the two books queries into one and populating both partners (allies) and resources from it is right, but it is always possible that I miss things.
The other change I put in is to have the userModel cache its results so that it isn't called multiple times in a short span. If you think 15 seconds is too long, I can adjust it downward.

PR #761, opened 2017-02-22
Fix tagline under instructor resources

RoyEJohnson:
insertHtml should insert an empty string when it gets an undefined value


RoyEJohnson:
[Fix tagline under Instructor Resources](https://trello.com/c/VeEbsBBm/161-fix-tagline-under-instructor-resources)

PR #770, opened 2017-02-28
Fix errata progress bar final point

RoyEJohnson:


RoyEJohnson:
[Errata tracking page: When in review, final point on progress bar should say "Corrected."](https://trello.com/c/ZNxkXxWs/190-errata-tracking-page-when-in-review-final-point-on-progress-bar-should-say-corrected)

PR #773, opened 2017-02-28
Fix blog article layout

RoyEJohnson:


RoyEJohnson:
[Blog article content divided into short columns](https://trello.com/c/n3xIfylE/195-blog-article-content-divided-into-short-columns)

PR #782, opened 2017-03-02
Cache page data

RoyEJohnson:
CMS page data is cached so returning to the page does not require another fetch.
Cached data self-purges in 15 minutes.

### DIFF
--- a/src/app/controllers/cms.js
+++ b/src/app/controllers/cms.js
@@ -1,4 +1,5 @@
 import settings from 'settings';
+import jsonDataCache from '~/helpers/json-data-cache';
 import {Controller} from 'superb';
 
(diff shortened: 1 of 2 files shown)


philschatz:
Don't browsers already do this (as long as the server sends the proper etag headers)?

edwoodward:
If we implement caching it should be server-side. I don't think this is needed on the frontend.

RoyEJohnson:
There's a certain amount of latency in switching between pages (say, between subjects and higher ed). If the data is cached rather than being re-pulled (even from browser cache), the switch is noticeably faster.

RoyEJohnson:
Confirmed that browser caching is turned off.

PR #784, opened 2017-03-06
Use CMS for order of books on Subjects page

RoyEJohnson:


RoyEJohnson:
[Control order of books on Subjects page](https://trello.com/c/0DP3KmTF/165-control-order-of-books-on-subjects-page)

PR #785, opened 2017-03-06
Fix errata confirmation page copy

RoyEJohnson:


RoyEJohnson:
[Errata form: email confirmation field needs different layout and behavior.](https://trello.com/c/CRfRbECT/181-errata-form-email-confirmation-field-needs-different-layout-and-behavior)

PR #787, opened 2017-03-06
Fix: pages that don't use the loader animation stay hidden

RoyEJohnson:
Remove page-hiding css from index.html
Wait for #main to get content before attaching header and footer
Determine whether page-loader is being used; if not, add page-loaded
class to prevent its running later


RoyEJohnson:
[[OSW2.design.08] Reimplement loading animation](https://trello.com/c/l0M9HYaC/14-osw2-design-08-reimplement-loading-animation)

PR #788, opened 2017-03-07
Do not auto-route staff/superuser to Give page from Get This Title

RoyEJohnson:


RoyEJohnson:
[Do not auto-route staff/admin to give page](https://trello.com/c/3W7zl4Xf/208-do-not-auto-route-staff-admin-to-give-page)

PR #800, opened 2017-03-10
Errata form: stop option from jumping between columns

RoyEJohnson:


RoyEJohnson:
[Fix alignment of descriptor text on errata form](https://trello.com/c/lsozwM4y/217-fix-alignment-of-descriptor-text-on-errata-form)

PR #806, opened 2017-03-16
Route invalid errata pages to 404

RoyEJohnson:


RoyEJohnson:
[Errata page should go to 404 page when given an invalid path](https://trello.com/c/tx2Undoq/245-errata-page-should-go-to-404-page-when-given-an-invalid-path)

PR #809, opened 2017-03-21
Fix: errata form labels overlap when they wrap

RoyEJohnson:


RoyEJohnson:
[BUG: Overlapping fields in errata when the screen size is reduced slightly.](https://trello.com/c/K0k01pYa/246-bug-overlapping-fields-in-errata-when-the-screen-size-is-reduced-slightly)

PR #815, opened 2017-03-23
Route Bulk Order and Contact confirmations correctly

RoyEJohnson:


RoyEJohnson:
[After receiving bulk-order, route to /confirmation/bulk-order](https://trello.com/c/LRWTpFuI/229-after-receiving-bulk-order-route-to-confirmation-bulk-order)

PR #816, opened 2017-03-24
Fix: Display instructor resources whose links are hidden

RoyEJohnson:


RoyEJohnson:
[Fix display of Instructor resources](https://trello.com/c/t3hSYSHh/253-fix-display-of-instructor-resources)

PR #818, opened 2017-03-28
Selected item is gray, unselected is white

RoyEJohnson:
Also made category-selector a subclass of radio-panel


RoyEJohnson:
[Update Selection Pattern on OS Web to Match Tutor](https://trello.com/c/krlOdPvT/255-update-selection-pattern-on-os-web-to-match-tutor)

PR #819, opened 2017-03-28
Fix: display book info on bulk-order

RoyEJohnson:


RoyEJohnson:
[BUG: Item, quantity, and list not displaying on bulk order form (QA/Dev)](https://trello.com/c/Z7tfWmyf/256-bug-item-quantity-and-list-not-displaying-on-bulk-order-form-qa-dev)

PR #820, opened 2017-03-30
Fix link color in support page header

RoyEJohnson:


RoyEJohnson:
[FAQ link on the Support page needs to change colors to be more visable.](https://trello.com/c/kDr8MOVA/258-faq-link-on-the-support-page-needs-to-change-colors-to-be-more-visable)

PR #821, opened 2017-03-30
Fix empty faculty resource link issue

RoyEJohnson:
Include credentials in slug fetches


RoyEJohnson:
[Red box around some instructor resources once you are signed in. Other instructor resources give you an access denied error even though you have instructor access.](https://trello.com/c/Icf7W21J/257-red-box-around-some-instructor-resources-once-you-are-signed-in-other-instructor-resources-give-you-an-access-denied-error-even-)

PR #834, opened 2017-04-06
Show books with comp copy available in comp copy form

RoyEJohnson:


RoyEJohnson:
[Add comp copy available checkbox to CMS, use for comp-copy form book list](https://trello.com/c/JUPGDIfZ/268-add-comp-copy-available-checkbox-to-cms-use-for-comp-copy-form-book-list)

PR #836, opened 2017-04-10
Catch analytics import failure

RoyEJohnson:


RoyEJohnson:
[Firefox does not recognize selection of None in multi-select](https://trello.com/c/UmdJwIcm/271-firefox-does-not-recognize-selection-of-none-in-multi-select)

PR #837, opened 2017-04-11
Fix position of filter button bar

RoyEJohnson:


RoyEJohnson:
[Fix position of filter button bar on Subjects and Partners pages](https://trello.com/c/hnuCWDhl/282-fix-position-of-filter-button-bar-on-subjects-and-partners-pages)

PR #842, opened 2017-04-18
Preselect book on Adoption form when clicking through from Details

RoyEJohnson:


RoyEJohnson:
[Book is not preselected on adoption form when clicking through from detail page](https://trello.com/c/NcDKLzpg/276-book-is-not-preselected-on-adoption-form-when-clicking-through-from-detail-page)

PR #845, opened 2017-04-19
TOC: treat "Thinking Ahead" as a non-numbered introduction

RoyEJohnson:


RoyEJohnson:
[Add Thinking Ahead as an intro module](https://trello.com/c/oKHt7wgc/274-add-thinking-ahead-as-an-intro-module)

PR #847, opened 2017-04-20
Fix passing book title from Errata to Adoption

RoyEJohnson:


RoyEJohnson:
[Book is not preselected on adoption form when clicking through from detail page](https://trello.com/c/NcDKLzpg/276-book-is-not-preselected-on-adoption-form-when-clicking-through-from-detail-page)

PR #851, opened 2017-04-20
Make minimum displayed donation $10

RoyEJohnson:


RoyEJohnson:
[Change lowest donation to $10](https://trello.com/c/eKisYpPe/290-change-lowest-donation-to-10)

edwoodward:
@philschatz We are getting less donations which the reason for the change. The old site had $10 as the default.

PR #870, opened 2017-05-01
Use label-value pairs for Adopt/Interest form roles

RoyEJohnson:


RoyEJohnson:
[BUG: Need separate value for Salesforce name on adopt/interest form roles](https://trello.com/c/mXmQ85IH/300-bug-need-separate-value-for-salesforce-name-on-adopt-interest-form-roles)

PR #873, opened 2017-05-02
More unit tests

RoyEJohnson:


philschatz:
o no, why did coverage decrease?

RoyEJohnson:
@philschatz Don't know. It also decreased by a fraction in the "Set Salesforce home" PR. In that case, it's probably because I added a line to an imported-but-untested area. In this case, maybe the new tests import too much untested code.

RoyEJohnson:
I also can't get it to pass tests on Travis that I can get it to pass locally.

package.json
```diff
@@ -73,6 +73,7 @@
       "settings": "<rootDir>/dist/settings",
       "fetch": "<rootDir>/jspm_packages/npm/whatwg-fetch@1.1.1/fetch.js"
     },
+    "coveragePathIgnorePatterns": ["/analytics", "/recordo", "/whatwg-fetch", "/router"],
```
philschatz: Could this just be the `jspm_modules` directory?

src/app/pages/subjects/subjects.js
```diff
@@ -31,12 +31,10 @@ export default class Subjects extends CMSPageController {
 
         shell.showLoader();
 
-        if (!$.isNode()) {
-            router.replaceState({
-                filter: this.categoryFromPath(),
-                path: pagePath
-            });
-        }
+        router.replaceState({
```
philschatz: ooh, no more `$.isNode()`? How'd that happen?

src/app/pages/subjects/subjects.js
```diff
@@ -55,15 +53,13 @@ export default class Subjects extends CMSPageController {
 
         this.bookViewer.filterSubjects(category);
 
-        if (!$.isNode()) {
-            const yTarget = history.state.y;
+        const yTarget = history.state.y;
```
philschatz: hmm, could/should this be `window.history` to clarify that this isn't a local variable?

RoyEJohnson: Could be, but there are many uses of history throughout the code base already.

package.json
RoyEJohnson: Yes, apparently. :)

test/setupFile.js
```diff
@@ -3,6 +3,7 @@ import 'babel-polyfill';
 jest.mock('~/components/shell/shell');
 
 global.fetch = require('jest-fetch-mock');
+global.regeneratorRuntime = regeneratorRuntime;
```
philschatz: OIC, cool!

And it looks like the line above could be switched to `require('fetch-vcr')` and you can record the HTTP responses from Django 😄

RoyEJohnson: That's an outdated change, but there is a different fetch package that could be swapped out.

PR #895, opened 2017-06-01
Remove hidden banner buttons from tab order

RoyEJohnson:
Also fix problem with Subjects menu popping open anytime a subjects
link is focused.


RoyEJohnson:
[Hidden banner buttons should not be keyboard-navigable](https://trello.com/c/qIDBzjAJ/337-hidden-banner-buttons-should-not-be-keyboard-navigable)

PR #901, opened 2017-06-12
Add tutor link and training wheel

RoyEJohnson:


RoyEJohnson:
[[OSW2.onboarding.11] - Tutor shortcut on OSWeb / As OpenStax, we want to add a Tutor link and training wheel on OSWeb, so that we can support Tutor onboarding](https://trello.com/c/TnZcgRQE/259-osw2onboarding11-tutor-shortcut-on-osweb-as-openstax-we-want-to-add-a-tutor-link-and-training-wheel-on-osweb-so-that-we-can-supp)

PR #902, opened 2017-06-12
Only transition banners when window is active

RoyEJohnson:


RoyEJohnson:
[Banner images getting overlaid in Safari](https://trello.com/c/aMzyC6AO/341-banner-images-getting-overlaid-in-safari)

PR #924, opened 2017-06-28
Hide Give sticky note after visiting give confirmation page

RoyEJohnson:


RoyEJohnson:
[BUG Give sticky does not disappear after visiting the give page](https://trello.com/c/P8Hn8uc4/369-bug-give-sticky-does-not-disappear-after-visiting-the-give-page)

PR #928, opened 2017-06-30
Fix "Got It" button in training wheel

RoyEJohnson:


RoyEJohnson:
[Hotfix: Got It button doesn't dismiss training wheel](https://trello.com/c/maxwXkiF/373-hotfix-got-it-button-doesnt-dismiss-training-wheel)

PR #941, opened 2017-08-04
Bulk order: Set initial quantity to 25, since min is 25

RoyEJohnson:
Though the field showed quantity of 25 because its min was 25, the actual quantity value was zero, and zero-quantity entries aren't included. Set initial quantity to 25.

### DIFF
--- a/src/app/pages/bulk-order/bulk-order.js
+++ b/src/app/pages/bulk-order/bulk-order.js
@@ -23,7 +23,7 @@ class OrderItems extends CMSPageController {
 
         this.model.orderItems = pages.map((p) => ({
             item: p.title,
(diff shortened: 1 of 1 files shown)


RoyEJohnson:
[Bulk Orders coming in with no book or quantity specified](https://trello.com/c/pVYnlD5E/400-bulk-orders-coming-in-with-no-book-or-quantity-specified)

edwoodward:
What about the book name not being sent? Will this correct that?

RoyEJohnson:
Yes, the book title and quantity go together. If the quantity is zero, it's not included. The fix is to make the quantity nonzero.

PR #951, opened 2017-09-12
Ensure users see book and not just give page

RoyEJohnson:


RoyEJohnson:
[Bug: Users trying to access OpenStax books on OS web only see the Give page](https://trello.com/c/iRljqTfE/441-bug-users-trying-to-access-openstax-books-on-os-web-only-see-the-give-page)

PR #952, opened 2017-09-14
Use footer copy from CMS

RoyEJohnson:


RoyEJohnson:
[Make footer copy editable](https://trello.com/c/RxtREAfl/334-make-footer-copy-editable)

PR #953, opened 2017-09-14
Debugging: log message when opening new window

RoyEJohnson:
Apparently when Firefox “blocks a pop-up” it does not return a null as
the window, which it does when window.open fails. Behavior is different
on dev than on local, so this is a testing PR.


RoyEJohnson:
[Bug: Users trying to access OpenStax books on OS web only see the Give page](https://trello.com/c/iRljqTfE/441-bug-users-trying-to-access-openstax-books-on-os-web-only-see-the-give-page)

PR #964, opened 2017-10-03
always use yarn

philschatz:
This is because deployment servers did not have yarn installed globally.

You can test this by doing: 

```
# Uninstall yarn. Depending on how you installed it, the uninstall is different:
brew uninstall yarn
npm uninstall -g yarn

./script/setup && ./script/test
```

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -8,9 +8,12 @@ source ./script/bootstrap || exit 1
 
 >&2 echo "==> Installing node packages"
 if [[ $(which yarn) ]]; then
+  >&2 echo "Using existing system version of yarn"
   yarn install || exit 2
 else
-  npm install || exit 2
+  >&2 echo "Installing yarn locally and then installing packages"
+  npm install yarn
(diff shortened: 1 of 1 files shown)


philschatz:
@RoyEJohnson Can you try deploying this before merging the code to verify that yarn is being installed locally?

philschatz:
I think that’s one of the ways it can be installed. I’m not sure about how best to install it. Would be nice to chat w/ ops about how best to do it. The options I see are:

1. `curl -o- -L https://yarnpkg.com/install.sh | bash || exit 1`
2. `if [[ $(which brew) ]]; then brew install yarn; else; sudo apt-get install yarn; fi`
3. `npm install --global yarn` (does not work when folks use `nvm`)
4. `npm install yarn` (annoyingly requires installing additional packages all the time)

philschatz:
Refs https://github.com/Connexions/cnx-recipes/pull/346

PR #969, opened 2017-10-09
Call sendToSalesforce after donation

RoyEJohnson:
Code was moved into a function, but the function wasn’t called.


RoyEJohnson:
[Donation Leads aren't showing up in SF anymore](https://trello.com/c/SNnCMVQ7/455-donation-leads-arent-showing-up-in-sf-anymore)

PR #970, opened 2017-10-10
Fix Salesforce form routing for non-production environments

RoyEJohnson:
Use the Tutor QA sandbox did

### DIFF
--- a/src/app/models/salesforce.js
+++ b/src/app/models/salesforce.js
@@ -33,7 +33,8 @@ function adoption(options) {
 const salesforce = {
     adoption,
     adoptionName,
-    salesforceHome: settings.testingEnvironment ? 'test.salesforce.com' : 'webto.salesforce.com'
+    salesforceHome: settings.testingEnvironment ? 'test.salesforce.com' : 'webto.salesforce.com',
+    oid: settings.testingEnvironment ? '00Dc0000003vv3Q' : '00DU0000000Kwch'
 };
 
 export default salesforce;
--- a/src/app/pages/contact/contact.js
+++ b/src/app/pages/contact/contact.js
@@ -48,8 +48,7 @@ export default class Contact extends SalesforceForm {
             classes: ['contact-page', 'page']
         };
         this.model = {
-            salesforceHome: salesforce.salesforceHome,
-            salesforceOid: salesforce.oid,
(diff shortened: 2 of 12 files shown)


RoyEJohnson:
[Modify Salesforce Web To Lead URL](https://trello.com/c/3g1zAYKm/430-modify-salesforce-web-to-lead-url)

src/app/models/salesforce.js
```diff
@@ -33,7 +33,8 @@ function adoption(options) {
 const salesforce = {
     adoption,
     adoptionName,
-    salesforceHome: settings.testingEnvironment ? 'test.salesforce.com' : 'webto.salesforce.com'
+    salesforceHome: settings.testingEnvironment ? 'test.salesforce.com' : 'webto.salesforce.com',
+    oid: settings.testingEnvironment ? '00Dc0000003vv3Q' : '00DU0000000Kwch'
```
philschatz: Could/should the settings file contain the salesforce id directly? This way, the settings.js file contains the specific keys should they need to change in the future. Similar to the webview settings.js file . Also:

1. if different environments needed to test a different salesforce id's they could without having to change this code
2. the testing salesforce id is not downloaded by every visitor to the website

src/app/pages/contact/contact.js
```diff
@@ -48,6 +48,7 @@ export default class Contact extends SalesforceForm {
         };
         this.model = {
             salesforceHome: salesforce.salesforceHome,
+            salesforceOid: salesforce.oid,
```
philschatz: If the settings suggestion above is used then I think this would just become `settings.SALESFORCE_OID` or something similar (and easy to grep to find all the occurrences of)

src/app/models/salesforce.js
RoyEJohnson: If we put it in settings.js, it becomes a devops change rather than a code change, and it has to be changed in each of the deployment environments.

src/app/pages/contact/contact.js
RoyEJohnson: I could give the models a `salesforce` member that is just the imported `salesforce` data. The template would then use `model.salesforce.salesforceHome` and `model.salesforce.oid`.

PR #971, opened 2017-10-11
replace github: with npm: package loading

philschatz:
I **think** this removes the remaining causes of rate limiting errors but I am unsure how to see which config (dev or not) is being used in this file. Maybe @RoyEJohnson would know?


philschatz:
Yep, I'm guessing they were only available on GitHub at the time os-webview was started (hence the `-alpha` version in the config) but have been added to npm since then.

That's why I'm hoping we can get https://greenkeeper.io running on all of our JS repos

PR #973, opened 2017-10-12
Try using Rollup

philschatz:
`jspm` has been causing deployment problems and is difficult to debug. Switching to rollup is an attempt to see how difficult it is to swap out the build system.

It still needs to copy several files into the `dist/` directory and will need to compile the sass files hopefully using the `sass` package (a dart reimplementation of the original ruby code which is easier to run from node)

# TODO

- [ ] re-use `gulp` to copy the non-JS and non-CSS files into the `dist/` directory
- [ ] get `./script/test` to run again
  - [ ] fix `jest` so it does not use `jspm`


philschatz:
I got the tests to pass and got rollup to build a JS file but:

1. the tests don’t use the built JS file (it only does unit testing)
2. we still need manual testing of the JS file (I haven’t done that yet)

philschatz:
closing in favor of #976

PR #975, opened 2017-10-20
Fix Polish "details" link and slug

RoyEJohnson:
Also relocate TOC-insertion code so it doesn’t depend on User info
being loaded.


RoyEJohnson:
["Details & Resources" link should be in Polish for the Polish book](https://trello.com/c/UbKgram7/481-details-resources-link-should-be-in-polish-for-the-polish-book)

PR #985, opened 2017-11-07
Hotfix: Fix analytics module's reliance on SystemJS

RoyEJohnson:


RoyEJohnson:
[Analytics is broken](https://trello.com/c/vVEwGCkL/504-analytics-is-broken)

src/app/helpers/analytics.js
```diff
@@ -16,22 +16,8 @@ class Analytics {
...
-                    () => {
-                        console.warn('GA FAILED TO IMPORT');
-                    }
-                ).catch((e) => {
-                    console.warn('Caught trying to import ga:', e);
-                });
-        } catch (e) {
-            console.warn('Caught SYSTEM JS nonsense', e);
-        }
+        window.ga('send', fields);
```
philschatz: Since the script is loaded `async` (in index.html) is `window.ga(...)` guaranteed to exist at this point?

Also, for folks that run an ad-blocker does `window.ga(...)` exist?

Could this log a message to stderr when `window.ga(...)` does not exist? That way devs can see that there was an attempt to send analytics?

Something like:

```js
if (typeof window.ga === 'function') {
  window.ga('send', fields);
} else {
  console.log('skipped-sending GoogleAnalytics send', fields);
}
```

RoyEJohnson: Graceful-failing code committed.

PR #993, opened 2017-11-21
Pass selected amount from Give sticky note

RoyEJohnson:


RoyEJohnson:
[BUG: Clicking on different amount of money on the Donate banner doesn't select the correct amount](https://trello.com/c/aV2bqGHI/586-bug-clicking-on-different-amount-of-money-on-the-donate-banner-doesnt-select-the-correct-amount)

PR #994, opened 2017-11-21
Stop webpack warnings

RoyEJohnson:
This doesn't get all the way to having webpack do all code generation, but it does reorganize and clean up some code. I'd like to get this into the master branch and do further work in a new branch.


RoyEJohnson:
[Revise FE build process for faster updates and cleaner build output](https://trello.com/c/MrtCvyYo/519-revise-fe-build-process-for-faster-updates-and-cleaner-build-output)

gulp/tasks/scripts.js
```diff
@@ -189,8 +188,6 @@ function compileScriptsBabel() {
     return gulp.src(`${config.src}/**/*.js`, {
         // since: gulp.lastRun('compileScriptsBabel')
     })
-    // .pipe(pi.sourcemaps.init({loadMaps: true}))
-    .pipe(pi.sourcemaps.init())
```
philschatz: Are JS & CSS sourcemaps being removed entirely or just to another place? At a glance I don't see them being re-added anywhere below. (maybe they're in `gulp/tasks/webpack.js`?)

I think sourcemaps should continue to be included because:

1. our code is opensource
2. it makes debugging production problems easier
3. it allows for code coverage if you run tests in a browser (like a11y tests)

gulp/tasks/webpack.js
```diff
@@ -0,0 +1,45 @@
...
+            rules: [
+                {
+                    test: /\.css$/,
+                    loader: 'ignore-loader'
+                }
+            ]
+        },
+        plugins: [
+            new webpack2.optimize.UglifyJsPlugin(),
+            new webpack2.optimize.MinChunkSizePlugin({minChunkSize: 16000})
```
philschatz: How was this size determined? Consider including a comment in the code so future-selves will know

src/app/pages/details/details-loaded-polish.html
```diff
@@ -54,7 +54,7 @@ <h2> Autorzy</h2>
             </section>
             <section class="floating-menu {model.bottom}">
                 <div class="box">
-                <h2>Weź tę książkę</h2>
+                <h2>Przejdź do książki</h2>
```
philschatz: Why did these change as a result of removing warnings?

src/app/pages/errata/errata.html
```diff
@@ -6,23 +6,7 @@ <h1 if="model.title" data-html="title()"></h1>
...
-                                  <i class="fa fa-cloud-download" aria-hidden="true"></i>
-                                  Download the latest release notes
-                              </a>
-                          </p>
-                          <p>
-                          Errata for OpenStax Tutor are updated at the end of each semester.
-                          </p>
-                          Have questions?  Please email <a href="mailto:errata@openstax.org">errata@openstax.org</a>
-                      </div>
+                    <div class="tooltip" data-html="errataHoverHtml"></div>
```
philschatz: Why did this change as a result of removing warnings?

philschatz:
It looks like this PR is based off of another PR and contains more than just changes to the warnings. Could it be marked as WIP and at least temporarily change the destination of the PR to be to the PR that it was based off of rather than `master`?

gulp/tasks/scripts.js
RoyEJohnson: Sourcemaps were not working after adding webpack as far as I was able to tell. Maybe I don't know what I'm looking for. Do you see evidence of sourcemaps on qa?

PR #997, opened 2017-11-27
Use errata hover language from CMS

RoyEJohnson:


RoyEJohnson:
[Add hover language on Errata list pages to CMS so we can add iBooks language](https://trello.com/c/ZkbcuBaG/505-add-hover-language-on-errata-list-pages-to-cms-so-we-can-add-ibooks-language)

PR #1002, opened 2017-11-30
Revise build system

RoyEJohnson:
This is a PR against `anticipated-merge`, a merge of `exclude-settings.js-from-bundle` and `master`. That shows which changes are actually new for purposes of code review.

### DIFF
--- a/gulp/tasks/scripts.js
+++ b/gulp/tasks/scripts.js
@@ -188,6 +188,8 @@ function compileScriptsBabel() {
     return gulp.src(`${config.src}/**/*.js`, {
         // since: gulp.lastRun('compileScriptsBabel')
     })
+    .pipe(pi.sourcemaps.init({loadMaps: true}))
+    .pipe(pi.sourcemaps.init())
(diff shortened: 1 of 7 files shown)


philschatz:
A fix for the sourcemaps is started in #1003

RoyEJohnson:
[Revise FE build process for faster updates and cleaner build output](https://trello.com/c/MrtCvyYo/519-revise-fe-build-process-for-faster-updates-and-cleaner-build-output)

RoyEJohnson:
@philschatz Is this PR ok for merge with the understanding that #1003 will follow? The card in my previous comment is not being closed out; I'm attaching 1003 to it as well.

philschatz:
In that case could the commits that remove the sourcemaps be removed from this PR? That way #1003 would not need to re-add them in.

RoyEJohnson:
There's no commit devoted to removing sourcemaps. That was part of code relocation and other revisions.

philschatz:
In that case can the code be re-added into this PR since it is not going the be removed?

PR #1005, opened 2017-12-04
Merge 1002 to master

RoyEJohnson:
Changes were reviewed and approved by Phil, but merged into a different branch; it needs to be merged to master now.

Drop sourcemaps, as they do not seem to be used
Remove redundant webpack step and config
Copy settings-example.js for dev
* Stop webpack warnings
* Separate webpack task so both scripts and templates can use it
Reload browser works for html updates as well as script updates
* Use webpack to minify code
Drop sourcemaps, as they do not seem to be used
Remove redundant webpack step and config
Copy settings-example.js for dev
Use global SETTINGS variable
Remove dead build scripts, keep fiddling with settings
* Add source map calls back in

### DIFF
--- a/gulp/tasks/scripts.js
+++ b/gulp/tasks/scripts.js
@@ -1,4 +1,3 @@
-const path = require('path');
 const gulp = require('gulp');
 const webpack = require('webpack-stream');
 const argv = require('yargs').argv;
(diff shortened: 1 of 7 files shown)


RoyEJohnson:
[Bundle and minify JS modules](https://trello.com/c/GZNrQIrv/536-bundle-and-minify-js-modules)

PR #1027, opened 2018-01-12
Let Get This Title popup menu overlap footer

RoyEJohnson:


RoyEJohnson:
[BUG: When we click on the "Place an Order" link the order now option is covered.](https://trello.com/c/E86vB3JF/554-bug-when-we-click-on-the-place-an-order-link-the-order-now-option-is-covered)

PR #1029, opened 2018-01-12
Go to top of page when loading blog article

RoyEJohnson:


RoyEJohnson:
[Bug: screen does not return to the top of the page when viewing more than one blog entry](https://trello.com/c/5zRXTc04/621-bug-screen-does-not-return-to-the-top-of-the-page-when-viewing-more-than-one-blog-entry)

PR #1034, opened 2018-01-16
Fix select widget's keyboard functionality

RoyEJohnson:


RoyEJohnson:
[BUG: "State" drop down in the Give form does not respond to alpha character input](https://trello.com/c/UlDYcxua/560-bug-state-drop-down-in-the-give-form-does-not-respond-to-alpha-character-input)

PR #1035, opened 2018-01-17
Stick Subjects filter to nav bar on mobile

RoyEJohnson:
Also adjust books-per-line at various widths


RoyEJohnson:
['Filter By' bar appears to be detached](https://trello.com/c/3u866Obc/644-filter-by-bar-appears-to-be-detached)

PR #1043, opened 2018-01-25
Keep "Get This Book" menu from hanging off screen

RoyEJohnson:
Increase height of screen at which menu is positioned statically.
Set max-height of floating menu and allow it to scroll if necessary (in
case  content makes it taller than we expect).


RoyEJohnson:
[BUG: "Get this book" can't scroll to sign up for info when "Place an order" is open on shorter screens.](https://trello.com/c/Wv8SWZ5F/555-bug-get-this-book-cant-scroll-to-sign-up-for-info-when-place-an-order-is-open-on-shorter-screens)

PR #1059, opened 2018-02-22
Add cache-busting parameters to CSS files and bundle.js

RoyEJohnson:
Add cache-busting parameters to CSS and bundle.js file requests


edwoodward:
@RoyEJohnson If the version is hardcoded, how will this work on the next release? It is manually updated?

PR #1065, opened 2018-03-05
Use v2 api for books and blog data

RoyEJohnson:


RoyEJohnson:
[Upgrade BE to Wagtail 2.0 - Upgrade FE to use API v2](https://trello.com/c/GSjQF0qV/624-upgrade-be-to-wagtail-20-upgrade-fe-to-use-api-v2)

PR #1066, opened 2018-03-06
Fix request for nonexistent math.svg

RoyEJohnson:
The generic css for the banners sets background image for .book and
.second-book to be ${subject}.svg. There is no math.svg. Although the
css overrides that generic css, we are still getting requests. I have
put in an if-directive to prevent the setting of the background image
in the generic css when subject is math.


RoyEJohnson:
[Bug: Math banner generates requests for nonexistent math.svg image](https://trello.com/c/RhNDjGVG/652-bug-math-banner-generates-requests-for-nonexistent-mathsvg-image)

PR #1075, opened 2018-03-15
Institutional partnership new

maherelaridi:


src/app/router.js
```diff
@@ -68,6 +69,7 @@ class AppRouter extends Router {
         this.route('give-confirmation', 'give');
         this.route('interest-confirmation', 'confirmation');
         this.route('higher-ed', 'home');
+        this.route('institutional-partnership-application', 'institutional-partnership');
```
RoyEJohnson: You shouldn't need this line. This section is for special routes. The entry on line 39 handles the regular routing.

src/app/pages/institutional-partnership/institutional-partnership.js
```diff
@@ -0,0 +1,35 @@
...
+        this.view = {
+            tag: 'main',
+            classes: ['institutional-page', 'page']
+        };
+    }
+
+    onLoaded() {
+        this.programTestimonial = this.el.querySelector('.testimonial-program');
+        this.applicationTestimonial = this.el.querySelector('.testimonial-application');
+        this.programTestimonial.classList.add('is-show');
```
RoyEJohnson: Apart from the top-level classes for the component, classes should be set in the view using state variables. In this case, you might have `this.whichTestimonialIsShowing` -- you'd probably choose a shorter name :) -- and your model would set variables according to it, like `this.model.testimonialShowClass = this.whichTestimonialIsShowing === 'application' ? 'is-show' : ''` and the template would have like `<div class="testimonial-application {model.testimonialShowClass}">`
If this doesn't make sense to you, put a TODO comment in here and I'll fix it up later since we're in a time crunch.

RoyEJohnson:
[[OSW2.partners.03] Implement Institutional Partner Application](https://trello.com/c/oIZDkXzs/628-osw2partners03-implement-institutional-partner-application)

PR #1077, opened 2018-03-16
Website refresh part 1

RoyEJohnson:
Updates to the new (still hidden) get this title, and comp copy modals on the new (still hidden) details page.


RoyEJohnson:
[Make Comp Copy form faculty-only - Update fields](https://trello.com/c/xelN9d8Y/614-make-comp-copy-form-faculty-only-update-fields)

PR #1078, opened 2018-03-19
Redirect back to current page after login

RoyEJohnson:
It doesn't look like logout handles redirect. It also looks like it doesn't reliably handle logging out. We may need changes in Accounts or the CMS (which passes thru the logout request).

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "http-server": "^0.10.0",
     "ignore-loader": "^0.1.2",
     "particles.js": "^2.0.0",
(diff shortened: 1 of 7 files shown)


RoyEJohnson:
[Bring users back to the page where they log in from.](https://trello.com/c/xSmXsQGU/655-bring-users-back-to-the-page-where-they-log-in-from)

RoyEJohnson:
[Logging out opens another tab](https://trello.com/c/mN6aFxzH/656-logging-out-opens-another-tab)

PR #1079, opened 2018-03-20
Preselect book on adoption form in Firefox

RoyEJohnson:
Set the property, not just the attribute


RoyEJohnson:
[FF: adoption form doesn't prefill selected book](https://trello.com/c/rOkjRMkt/659-ff-adoption-form-doesnt-prefill-selected-book)

PR #1080, opened 2018-03-20
Show error message on give form when country is empty

RoyEJohnson:
Remove self-closing tags


RoyEJohnson:
[Bug: country field on the donation form doesn't show an error message that it is required](https://trello.com/c/n2XvXYEX/660-bug-country-field-on-the-donation-form-doesnt-show-an-error-message-that-it-is-required)

PR #1082, opened 2018-03-20
Update Errata detail field to match BE max length

mwharrison:


mwharrison:
[Expand field for Errata detail](https://trello.com/c/zMBooekP/661-expand-field-for-errata-detail)

PR #1083, opened 2018-03-20
Enable new comp copy form

RoyEJohnson:


RoyEJohnson:
[Make Comp Copy form faculty-only - Update fields](https://trello.com/c/xelN9d8Y/614-make-comp-copy-form-faculty-only-update-fields)

PR #1091, opened 2018-04-12
Added nonInteraction flag to role

edwoodward:


edwoodward:
[OSWeb reporting incorrect bounce rate in GA](https://trello.com/c/rKLZ55NW/634-osweb-reporting-incorrect-bounce-rate-in-ga)

PR #1116, opened 2018-05-21
Stop double-counting on page load

RoyEJohnson:


RoyEJohnson:
[BUG: "Give Now" sticky is visible for frequent visitors](https://trello.com/c/eEY58LCZ/669-bug-give-now-sticky-is-visible-for-frequent-visitors)

PR #1119, opened 2018-05-24
Fix blog page to use news/ prefix for CMS but not for page URLs

RoyEJohnson:


RoyEJohnson:
[Bug: new blog URLs are inconsistent](https://trello.com/c/pXnwAhVs/679-bug-new-blog-urls-are-inconsistent)

PR #1130, opened 2018-06-11
Set return-to address correctly in Resources boxes

RoyEJohnson:


RoyEJohnson:
[Bug: logging in while on a book details page returns the user to book details instead of student/instructor resources](https://trello.com/c/z9knL65x/695-bug-logging-in-while-on-a-book-details-page-returns-the-user-to-book-details-instead-of-student-instructor-resources)

PR #1133, opened 2018-06-13
Fix Press page layout issues

RoyEJohnson:


RoyEJohnson:
[Bug: Press page probs](https://trello.com/c/MgszkTBr/827-bug-press-page-probs)

PR #1135, opened 2018-06-14
Fix experts issues on Press page

RoyEJohnson:


RoyEJohnson:
[Bug: Press page probs](https://trello.com/c/MgszkTBr/738-bug-press-page-probs)

PR #1139, opened 2018-06-19
Give be gone

edwoodward:


edwoodward:
[Stop navigating to Give page when the user goes to webview.](https://trello.com/c/1BhRGEvG/742-stop-navigating-to-give-page-when-the-user-goes-to-webview)

PR #1140, opened 2018-06-19
Fix Press Kit link on Press page

RoyEJohnson:


RoyEJohnson:
[Cannot download press kit](https://trello.com/c/l7PV0DXJ/743-cannot-download-press-kit)

PR #1141, opened 2018-06-19
New about us page

RoyEJohnson:
This creates the new About page at /about-new
The old About page is still active at /about


RoyEJohnson:
[[OSW2.refresh.17] Implement "About us"](https://trello.com/c/tUa21OJ9/736-osw2refresh17-implement-about-us)

PR #1142, opened 2018-06-21
Fix form links for Polish books

RoyEJohnson:


RoyEJohnson:
[Bug: Polish book links for adoption, interest, and errata are wrong](https://trello.com/c/jFEovAnv/748-bug-polish-book-links-for-adoption-interest-and-errata-are-wrong)

PR #1150, opened 2018-07-06
Do not 404 on Errata form when book parameter is wrong

RoyEJohnson:
Also fix title on Adoption form


RoyEJohnson:
[Bug: Errata form fails if book parameter is incorrect](https://trello.com/c/WaJnNMMt/767-bug-errata-form-fails-if-book-parameter-is-incorrect)

PR #1164, opened 2018-07-13
Fix Tutor training wheel disappears if page scrolls

RoyEJohnson:


RoyEJohnson:
[OSweb disabled if we scroll down without making selection in the Openstax Tutor Beta pop-up](https://trello.com/c/d7Y9wt3I/782-osweb-disabled-if-we-scroll-down-without-making-selection-in-the-openstax-tutor-beta-pop-up)

PR #1180, opened 2018-08-02
Use Bookstore Content for print order dialog

RoyEJohnson:


RoyEJohnson:
[Put print options button text in the CMS](https://trello.com/c/hVhl045L/779-put-print-options-button-text-in-the-cms)

PR #1183, opened 2018-08-06
Add new Business category to Subjects

RoyEJohnson:
It will only display categories that have members


RoyEJohnson:
[Add new Business subject to openstax.org](https://trello.com/c/7VEyD24e/806-add-new-business-subject-to-openstaxorg)

PR #1192, opened 2018-08-09
Use new errata call parameters

RoyEJohnson:


RoyEJohnson:
[CHANGE REQUEST: Archived errata should still be available at their direct URL](https://trello.com/c/Wvia36uE/808-change-request-archived-errata-should-still-be-available-at-their-direct-url)

PR #1202, opened 2018-08-17
Get user data from Accounts rather than Wagtail

RoyEJohnson:


RoyEJohnson:
[Use accounts.openstax for all user data (instead of the CMS)](https://trello.com/c/HMglYE1M/820-use-accountsopenstax-for-all-user-data-instead-of-the-cms)

RoyEJohnson:
[Bug: Changes to first name in account profile is not updating osweb](https://trello.com/c/RqBpnHGG/847-bug-changes-to-first-name-in-account-profile-is-not-updating-osweb)

PR #1212, opened 2018-08-27
Fix Partners page subject bar position

RoyEJohnson:


RoyEJohnson:
[Bug: Partners screen - subjects bar is mis-aligned](https://trello.com/c/Ui3QSqnl/911-bug-partners-screen-subjects-bar-is-mis-aligned)

PR #1214, opened 2018-08-29
Have at most one main tag per page

RoyEJohnson:
Or none.


RoyEJohnson:
[Bug: There should be at most one <main> tag per page](https://trello.com/c/CdXHNRfZ/841-bug-there-should-be-at-most-one-main-tag-per-page)

PR #1217, opened 2018-08-30
Change submitted_by to submitted_by_account_id for Errata form

RoyEJohnson:


RoyEJohnson:
[Change Errata to use user account data from accounts.openstax.org, instead of local CMS user accounts](https://trello.com/c/8Bjx0JNi/844-change-errata-to-use-user-account-data-from-accountsopenstaxorg-instead-of-local-cms-user-accounts)

PR #1220, opened 2018-09-04
Create Subjects menu from CMS data

RoyEJohnson:
All requests for api/books now use one promise
subjectCategories promise filters subject categories to include only the ones for which we have books


RoyEJohnson:
[Add new Business subject to openstax.org](https://trello.com/c/7VEyD24e/806-add-new-business-subject-to-openstaxorg)

PR #1221, opened 2018-09-05
Close "What we do" menu on resize

RoyEJohnson:


RoyEJohnson:
[Bug: Menu options look wierd in ipad - when we toggle between horizontal vs vertical view](https://trello.com/c/W8Vb0smM/843-bug-menu-options-look-wierd-in-ipad-when-we-toggle-between-horizontal-vs-vertical-view)

PR #1231, opened 2018-09-13
Allow cancel from comp-copy dialog

RoyEJohnson:


RoyEJohnson:
[Bug: "We measure" is covered by the pinned nav bar when the window is just above the 760 transition width and text is lost at the bottom](https://trello.com/c/B13H33hy/937-bug-we-measure-is-covered-by-the-pinned-nav-bar-when-the-window-is-just-above-the-760-transition-width-and-text-is-lost-at-the-b)

PR #1232, opened 2018-09-13
Fix Technology page resource links

RoyEJohnson:


RoyEJohnson:
[Bug: tech links aren't valid now that the book details page is split into tabs](https://trello.com/c/VxkASo3d/938-bug-tech-links-arent-valid-now-that-the-book-details-page-is-split-into-tabs)

PR #1235, opened 2018-09-14
Dynamically load blog articles

RoyEJohnson:
And begin excerpt with the first `<p>`


RoyEJohnson:
[Pull from first <p> text in blog preview](https://trello.com/c/oW7jqCUt/872-pull-from-first-p-text-in-blog-preview)

PR #1236, opened 2018-09-17
Fix excessive polling for Tutor

RoyEJohnson:
Was searching for the wrong value in groups list (value changed when we started using Accounts directly to generate user info)


RoyEJohnson:
[Tutor training wheel pops up when logged in randomly](https://trello.com/c/3Hc5bGVl/874-tutor-training-wheel-pops-up-when-logged-in-randomly)

PR #1237, opened 2018-09-18
Allow link text to wrap in blog articles

RoyEJohnson:
Also update version and remove no-longer-needed-classList CDN


RoyEJohnson:
[Linked text in blog won't wrap](https://trello.com/c/0GJso73x/908-linked-text-in-blog-wont-wrap)

PR #1251, opened 2018-10-11
Update CMS slugs to match page URI

mwharrison:


src/app/pages/subjects/subjects.js
```diff
@@ -17,7 +17,7 @@ export default class Subjects extends CMSPageController {
         'decide if they\'re right for your course.';
 
     init() {
-        this.slug = 'books';
+        this.slug = 'subjects';
```
RoyEJohnson: Will the slug "books" still work? Because it is used in `src/app/models/books.js`

PR #1252, opened 2018-10-16
Only include "live" books in Salesforce book list

RoyEJohnson:
Adoption and interest form show covers for old books because the Salesforce book list does not distinguish between editions and uses the data from the first entry for (say) Biology that comes from the CMS. Filter out entries whose `book_state` is not "live" to avoid using obsolete information.


RoyEJohnson:
[Bug: Adoption and interest forms showing old-edition book covers](https://trello.com/c/U6eyxEtG/925-bug-adoption-and-interest-forms-showing-old-edition-book-covers)

PR #1262, opened 2018-10-29
replace literal with a space

gregfitch:
the HTML space was being inserted as a literal


gregfitch:
https://github.com/openstax/os-automation/issues/164
(forgot to include this earlier...)

PR #1277, opened 2018-11-13
Clean up dist directory

RoyEJohnson:


script/build
```diff
@@ -3,6 +3,7 @@ cd "$(dirname "$0")/.." || exit 111
 source ./script/bootstrap || exit 111
 
 try npm run-script build
+/bin/rm -r ./dist/app ./dist/settings*.js*
```
philschatz:
```suggestion
do_progress "Removing temporary build files" \
  rm -r ./dist/app ./dist/settings*.js*
```

Using `do_progress` allows:

- readable text to show that something is happening
- if it fails, a detailed error message shows which line failed

RoyEJohnson: Seems to work.

PR #1338, opened 2019-02-12
Load Analytics first

RoyEJohnson:
So it will be available to send page view information.


RoyEJohnson:
TO VERIFY: Open the console before loading the site.
When the page loads, the console should _not_ display
`analytics.js:27 Google Analytics not loaded; did not send {hitType: "pageview", page: "/", location: "https://cms-dev.openstax.org/"}`

PR #1339, opened 2019-02-13
Send correct roles to Salesforce from Adoption and Interest forms

RoyEJohnson:
Instead of just "Faculty"


RoyEJohnson:
@mwharrison  Could you merge this, please?
@Malar-Natarajan Yes, it would be good if this could be in the current release.

PR #1386, opened 2019-04-23
Hash generated script and style files

nathanstitt:
Adds a content segment to the script and style files and references them by their name.

I **think** this also fixes a few bugs:
  * The `src/index.html` file had hard-coded versions used as the query parameter, so the original regex in html.js would never match since the placeholders were not present
  * The 'humans.js` grunt task would overwrite `dist/index.html` with `src/index.html`.  So even if the `html.js`  substation had worked, it would have been overwritten.

Also while I'm personally a big fan of 4 space indent, my editor was setup to use 2 space for all OX projects.  I added an `.editorconfig` file so that it'd behave.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,16 +1,17 @@
 const {series, parallel, task} = require('gulp');
 const allTheThings = require('require-dir')('./gulp/tasks');
 const {
-    clean, copy, development, production, favicon, html, scsslint, styles,
-    humans, images, eslint, scripts, settings, precache, templates, webpack,
-    watch
+  clean, copy, development, production, favicon, scsslint, styles,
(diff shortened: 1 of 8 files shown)


gulpfile.js
```diff
@@ -1,16 +1,17 @@
...
+  humans, images, eslint, scripts, settings, precache, templates, webpack,
+  watch, devHTML, distHTML
 } = Object.assign({}, ...Object.values(allTheThings));
 
+
 const beforeWebpack = series(
     clean,
     parallel(
         copy,
-        series(favicon, html),
```
jpslav: @nathanstitt what happened here?

nathanstitt: there used to be only one `html` step, but I replaced it with `devHTML` and `distHTML`, the dist variant adds the hashes.

PR #1392, opened 2019-04-26
Serve fonts from configured directory

nathanstitt:
Enables them to also be cached by cloudfront


nathanstitt:
@RoyEJohnson hold off on review/merge of this, am investigating why `dist/styles/fonts.css` no longer contains the font definitions like it used to

PR #1417, opened 2019-05-15
Route oxauth properly (2nd try)

RoyEJohnson:
I think the problem is that I removed a needed `resolve` call along with some debugging code.

### DIFF
--- a/src/app/helpers/link.js
+++ b/src/app/helpers/link.js
@@ -2,7 +2,7 @@ import settings from 'settings';
 
 const EXTERNAL = /^((f|ht)tps?:)?\/\//;
 const ABSOLUTE_OPENSTAX = new RegExp(
-    `(https?://openstax.org|${settings.apiOrigin})(?!/accounts|/oxauth|/blog-feed|/support)`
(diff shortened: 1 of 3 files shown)


src/app/helpers/link.js
```diff
@@ -2,7 +2,7 @@ import settings from 'settings';
 
 const EXTERNAL = /^((f|ht)tps?:)?\/\//;
 const ABSOLUTE_OPENSTAX = new RegExp(
-    `(https?://openstax.org|${window.location.origin})(?!/accounts|/oxauth|/blog-feed|/support)`
+    `(https?://openstax.org|${window.location.origin})/(?!accounts|oxauth|blog-feed|support)`
```
mwharrison: @RoyEJohnson - will this work on dev and QA with https://openstax.org hardcoded in?

RoyEJohnson: Yes, it's openstax.org _or_ window.location.origin. It accounts for some links that are to openstax.org in the CMS data. The usual case is window.location.origin.

PR #1426, opened 2019-05-23
Adds line break for book contents.

berkalpyakici:
Adds line break for book contents as requested.

### DIFF
--- a/src/app/pages/details/details.scss
+++ b/src/app/pages/details/details.scss
@@ -49,7 +49,7 @@
 
     .loc-summary-text {
         p {
-            padding-bottom: 8px;
+            padding-bottom: 0.8rem;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/app/pages/details/details.scss
```diff
@@ -47,6 +47,12 @@
         color: text-color(white);
     }
 
+    .loc-summary-text {
+        p {
+            padding-bottom: 8px;
```
RoyEJohnson: I expect this to fail lint. We use `rem`s instead of `px` for better cross-device/scaling behavior. The equivalent to `8px` is `0.8rem`.

RoyEJohnson: Bad news for me: lint doesn't complain about it, due to changing to sasslint, I presume.

berkalpyakici: Alright, I'll change it to 0.8rem.

PR #1433, opened 2019-05-31
Mapbox Redo. Fixes cards that are related to global-impact map.

berkalpyakici:
This introduces two new settings to settings.json.

mapboxStyle should be an environment setting that is different for all environments.

### DIFF
--- a/src/app/pages/global-reach/map/map.js
+++ b/src/app/pages/global-reach/map/map.js
@@ -221,6 +221,7 @@ export default class Map1 extends Controller {
             closeButton: false,
             closeOnClick: false
         });
+
         mtooltip.setLngLat([dField.long, dField.lat]);
         mtooltip.setHTML(`<b>${iName}</b><br>${pCity === null ? '': pCity}${pCity !== null && pState !== null ? ', ': ''}${pState === null ? '' : pState}`);
         mtooltip.addTo(this.mapObject);
@@ -433,21 +434,21 @@ export default class Map1 extends Controller {
         }
     }
     makeFitBoundsArray(dataArray) {
-        var latlngArray = dataArray.map((e) => {
+        let latlngArray = dataArray.map((e) => {
             return [Number.parseFloat(e.fields.long), Number.parseFloat(e.fields.lat)];
         });
 
-        latlngArray = latlngArray.filter(i => i[0] !== 0 && i[1] !== 0);
+        latlngArray = latlngArray.filter((i) => (i[0] !== 0 && i[1] !== 0));
 
         const bounds = latlngArray.reduce((bound, coord) => {
             return bound.extend(coord);
         }, new mapboxgl.LngLatBounds(latlngArray[0], latlngArray[0]));
 
-        
-        if('_sw' in bounds && '_ne' in bounds) {
+
+        if ('_sw' in bounds && '_ne' in bounds) {
             const setBound = [[bounds._sw.lng, bounds._sw.lat], [bounds._ne.lng, bounds._ne.lat]];
 
-            if(latlngArray.length > 1){
+            if (latlngArray.length > 1) {
                 this.mapObject.fitBounds(setBound, {
                     padding: 100
                 });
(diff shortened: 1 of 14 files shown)


src/app/pages/global-reach/map/map.js
```diff
@@ -211,6 +212,21 @@ export default class Map1 extends Controller {
         document.getElementById('back-result-div').setAttribute('style', 'display: none;');
         document.getElementById('back-detail-div').setAttribute('style', 'display: none;');
     }
+    generateCityState(city, state) {
```
RoyEJohnson: Could write as
```
function generateCityState(...args) {
    return args.filter((a) => a).join(', ');
}
```
:) (but it is not necessary that you do so)

src/app/pages/global-reach/map/map.js
```diff
@@ -428,19 +450,31 @@ export default class Map1 extends Controller {
         }
     }
     makeFitBoundsArray(dataArray) {
-        const latlngArray = dataArray.map((e) => {
-            return [Number.parseFloat(e.fields.lat), Number.parseFloat(e.fields.long)];
+        let latlngArray = dataArray.map((e) => {
+            return [Number.parseFloat(e.fields.long), Number.parseFloat(e.fields.lat)];
         });
 
+        latlngArray = latlngArray.filter((i) => (i[0] !== 0 && i[1] !== 0));
```
RoyEJohnson: Is this logic correct, or should it be or? This filters out if either one is zero. Also, you could chain this onto the previous statement (and have it go back to `const`).

src/app/pages/global-reach/map/mapdropdown.html
```diff
@@ -7,7 +7,8 @@
             <div class="toggle-on-off" id="show-{element.pk}" data-toggle="show">
                 <div class="list-data">
                     <div class="list-heading">{element.fields.name}</div>
-                    <div class="list-address">{element.fields.physical_city}{element.fields.physical_state_province === null ? '' : ', '+element.fields.physical_state_province}</div>
+                    <div class="list-address">{element.fields.physical_city === null ? '' : element.fields.physical_city}{element.fields.physical_city !== null && element.fields.physical_state_province !== null ? ',' : ''}
```
RoyEJohnson: You could make `generateCityState` available in the model, and call it here.

src/app/pages/global-reach/map/mapdropdown.js
```diff
@@ -112,19 +113,40 @@ export default class Mapdropdown extends Controller {
         document.getElementById('back-search-div').setAttribute('style', 'display: none;');
         document.getElementById('back-result-div').setAttribute('style', 'display: block;');
     }
+    generateCityState(city, state) {
```
RoyEJohnson: This is duplicated from `map.js`. It really doesn't need to be a method at all; it's just a helper function. It can be defined as a separate exportable in `map.js` and then imported like
```
import {generateCityState} from './map.js';
```
here.

src/app/pages/global-reach/map/schoolinfo-head.html
```diff
@@ -1,4 +1,6 @@
 <template args="model"></template>
     <div class="list-heading">{model.name}</div>
-    <div class="list-address">{model.physical_city}{model.physical_state_province === null ? '' : ', '+model.physical_state_province}</div>
+    <div class="list-address">
+            {model.physical_city === null ? '' : model.physical_city}{model.physical_city !== null && model.physical_state_province !== null ? ',' : ''}
```
RoyEJohnson: As before, this is a lot of logic to include in the view. It should be a function call. You can import `generateCityState` for this module, too.

src/app/pages/global-reach/map/map.js
berkalpyakici: We want to filter out if both are zero and keep the rest. I chained this and moved it back to constant.

RoyEJohnson: Here you are _including_ where _both_ are non-zero, so filtering out where at least one is zero. If you change `&&` to `||`, you will include with at least one is non-zero and only filter out where both are zero.

PR #1440, opened 2019-06-05
Increased padding and removed empty <p> and all <br> from descriptions.

berkalpyakici:


src/app/pages/details/details.js
```diff
@@ -140,7 +140,7 @@ export default class Details extends BaseClass {
             const model = {
                 bookInfo: this.pageData,
                 bookState: this.pageData.book_state,
-                description: this.pageData.description,
+                description: this.pageData.description.replace(/[<]br[^>]*[>]/gi, '').replace('<p></p>', ''),
```
RoyEJohnson: As a rule, avoid parsing HTML. One way around that is to create a `div`, and use `querySelector` to find and remove the elements you want to remove.

I think in this case we can just use CSS to hide the offending bits. In the `[data-html="description"]` section, try something like
```
p:empty,
br {
    display: none;
}
```

berkalpyakici: Thanks, this is was really helpful!

PR #1446, opened 2019-06-06
Ensure usermodel is loaded before page slug

RoyEJohnson:
Now that book details depend on the user (and other requests may, in the future), ensure user info is loaded before any slug.


RoyEJohnson:
Loading userinfo is not the same as setting the cookie

PR #1455, opened 2019-06-17
Update blog pages so images render

RoyEJohnson:
Promoted blog-search to replace blog
Updated to hide search bar for now, so it's just like the old blog pages, but with better structure and working images.


RoyEJohnson:
@berkalpyakici This is a pretty big one, but I'm hoping you have some available time to go through it. I have tried to write clear, sensible code. If you find it hard to understand what's going on, leave comments.

src/app/pages/blog/article-summary/article-summary.scss
```diff
@@ -29,6 +29,11 @@
 
     .article-headline {
         margin: 0;
+
+        a:not(:hover) {
```
berkalpyakici: Correct me if I'm wrong, but is there a better way to catch this state instead of using :not?

src/app/pages/blog/blog.js
```diff
@@ -3,61 +3,161 @@ import settings from 'settings';
...
+        const slugs = Reflect.ownKeys(articles)
+            .filter((k) => test(k, articles[k]))
+            .sort((a, b) => articles[a].date < articles[b].date ? 1 : -1);
+
+        return slugs.map((s) => blurbModel(s, this.pageData.articles[s]));
+    }
+
+    buildDefaultPage() {
+        const fa = new FeaturedArticle(this.featuredArticleOptions);
+        // const sb = new SearchBar({
```
berkalpyakici: Do we need the commented out part?

src/app/pages/blog/blog.js
```diff
@@ -3,61 +3,161 @@ import settings from 'settings';
...
+import UpdateBox from './update-box/update-box';
+import FeaturedArticle from './pinned-article/pinned-article';
+import MoreStories from './more-stories/more-stories';
+import {blurbModel} from './article-summary/article-summary';
+import Article from './article/article';
+import DisqusForm from './disqus-form/disqus-form';
 
+function slugsSortedByArticleDate(articles) {
+    return Object.keys(articles)
+        .sort((a, b) => {
```
berkalpyakici: Can this be reduced to `return articles[a].date < articles[b].date ? 1 : -1;`

src/app/pages/blog/blog.js
```diff
@@ -3,61 +3,161 @@ import settings from 'settings';
...
-                    this.previousSlug = slugMatch[1];
-                    this.setTimers();
-                    this.setCanonicalLink(slugMatch[0], this.canonicalLink);
-                }
+        });
+        const ms = new MoreStories({
+            articles: this.moreStoriesOptions((_, articleEntry) => !articleEntry.pin_to_top)
+        });
+
+        // sb.on('value', (searchParam) => {
```
berkalpyakici: Same here

src/app/pages/blog/blog.js
```diff
@@ -3,61 +3,161 @@ import settings from 'settings';
...
+        const ub = new UpdateBox({
+            model: {
+                rssUrl: `${settings.apiOrigin}/blog-feed/rss/`
+            }
+        });
+        const ms = new MoreStories({
+            articles: this.moreStoriesOptions((slug, _) => slug !== this.articleSlug)
+        });
+
+        this.regions.self.attach(a);
```
berkalpyakici: Can we just do
```
    this.regions.self.attach(new Article({
            slug: `news/${this.articleSlug}`
        }));
```

src/app/pages/blog/article-summary/article-summary.scss
RoyEJohnson: I did it this way because I wanted the default behavior for hover. If I specified this with just `a {}`, then I would have had to specify `a:hover {}` explicitly. This is the easier way to know I got what I wanted.

src/app/pages/blog/blog.js
RoyEJohnson: It can. I'll do that.

src/app/pages/blog/blog.js
RoyEJohnson: This is a future feature. I'll add a comment to the comments. :)

src/app/pages/blog/blog.js
RoyEJohnson: I'm a little undecided about readability of inlining the component creation, but it's not really bad. Reworked.

berkalpyakici: I just assumed that if we don't  need to access to a in the future, we wouldn't need to declare a new variable for that.

PR #1459, opened 2019-06-19
Fix login problem and turn off Hero Journey page

RoyEJohnson:
Put up a placeholder notice, but keep all the old code available to be turned back on.


mwharrison:
Approved - but should this be something we do in the CMS? Is this something that will need to be turned on and off?

PR #1465, opened 2019-06-24
Handle different-sizes of blog article images

RoyEJohnson:
If the alignment is "left" and not on phone, the image will be 50% of the text width and floated left. Similarly, "right" will float right.
Otherwise, the image will be up to the full width of the text.


mwharrison:
@RoyEJohnson - are you saying we should not have the two sizes in the blog post body?

PR #1477, opened 2019-07-01
Refactor map pages

RoyEJohnson:
Also covers issue #409: Some community colleges show up for college/university and vice-versa.


src/app/helpers/map-api.js
```diff
@@ -0,0 +1,136 @@
...
+        this.setBounds();
+        this.tooltip.remove();
+    }
+
+    showTooltip(schoolInfo, flyThere) {
+        if (!hasLngLat(schoolInfo)) {
+            return;
+        }
+        this.tooltip.setLngLat(schoolInfo.lngLat);
+        this.tooltip.setHTML(`<b>${schoolInfo.fields.name}</b><br>${schoolInfo.cityState}`);
```
berkalpyakici: Here if `schoolInfo.cityState` is `null`, then we probably don't want to have a line break (as the padding on the bottom may seem larger)

src/app/pages/global-reach/global-reach.js
```diff
@@ -31,82 +18,49 @@ const BaseClass = componentType(spec, canonicalLinkMixin, loaderMixin);
...
+                    imageUrl: obj.image.image,
+                    uText1: obj.number,
+                    uText2: obj.unit,
+                    lowerText: obj.description
                 }
-            }));
-        });
+            ))
+        }));
+        this.regions.self.append(new Studentinfo({
```
berkalpyakici: This model can use documentation for the fields.

src/app/models/querySchools.js
```diff
@@ -0,0 +1,73 @@
...
+    return fetch(`${searchPath}?${searchParameters}`)
+        .then((response) => response.json())
+        .then((response) =>
+            response.length > MAX_SEARCH_RESULTS ? [] :
+                response.map(augmentInfo)
+                    .sort((a, b) => a.fields.name.toLowerCase() < b.fields.name.toLowerCase() ? -1 : 1)
+        );
+}
+
+export default function (queryString, selectedFilters) {
```
berkalpyakici: This function can also use documentation (like indicating that is is communicating with CMS for API requests).

src/app/helpers/controller/bus-mixin.js
```diff
@@ -55,9 +55,32 @@ export class Bus {
 */
 export default (superclass) => class extends superclass {
 
+    constructor(...args) {
```
berkalpyakici: This is probably the only file that I had hard time reading (possibly because it's a bus)

src/app/helpers/map-api.js
RoyEJohnson: If `cityState` is null, will there be a point on the map? (Regardless, I made the fix.)

src/app/helpers/controller/bus-mixin.js
RoyEJohnson: I mis-documented this as I was writing it, which is likely one reason it was hard to read. I have given it correct documentation.

src/app/pages/global-reach/global-reach.js
RoyEJohnson: It wasn't clear to me how this was much different from any other section, but I gave it a shot.

src/app/helpers/map-api.js
berkalpyakici: Yeah, sometimes, we have the coordinates but not the name of the city/country.

PR #1487, opened 2019-07-15
Add user id tracking to Google Analytics

RoyEJohnson:
Added the call. Will need to verify that it's happening in dev.


mwharrison:
@RoyEJohnson @allisonhowell - we're okay to do this without the cookie policy in place?

PR #1529, opened 2019-08-13
Enable userid tracking for non-GDPR countries

RoyEJohnson:
I have set this up to look for JP's proposed `is_gdpr_location` and to not show the notice or set the tracking cookie unless that is explicitly false. That unblocks this and leaves the cookie/notice inactive until Accounts is serving that new flag.

### DIFF
--- a/src/app/components/dialog/dialog.scss
+++ b/src/app/components/dialog/dialog.scss
@@ -34,7 +34,7 @@ dialog {
         @include wider-than($phone-max) {
             align-items: center;
             color: inherit;
-            padding: 2.5rem 3rem;
+            padding: 1.5rem 3rem;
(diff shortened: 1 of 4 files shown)


TomWoodward:
should we make a small npm module for the country lookup so that i don't have to copy/paste your country dictionary?

RoyEJohnson:
> should we make a small npm module for the country lookup so that i don't have to copy/paste your country dictionary?

That sounds reasonable. I don't know how to go about it, though.

RoyEJohnson:
> pretty sure this will not work in production because of the https > http getting blocked

I have rewritten this to look for the expected flag in accounts rather than making the country check call.

PR #1534, opened 2019-08-22
Make new rover page live

RoyEJohnson:
Out with the old, in with the new


src/app/pages/rover-by-openstax/rover-by-openstax.js
```diff
@@ -110,59 +98,60 @@ export default class RoverRedesign extends BaseClass {
...
                         video: c.video
-                    }))
+                    })),
+                    moreText: 'See more resources',
+                    moreUrl: '/general/rover-onboarding'
                 }
             }),
-            faqSection({
+            new OfficeHours({
+                model: {
```
mwharrison: does this need a field in the CMS?

PR #1547, opened 2019-09-03
Fix local links opening in new tabs

RoyEJohnson:
Test for external URL needed to exclude OpenStax addresses.
(Formerly OS domain was pre-stripped, but don't want to do that with the new umbrella domain.)


RoyEJohnson:
@mwharrison This is a fix for the current release candidate. Brooke has verified it, so its card is in ready to deploy. It just needs to get into master as well.

PR #1548, opened 2019-09-04
Set "Sign up" link on Rover page

RoyEJohnson:
Fix for release


RoyEJohnson:
> candidate for the CMS?

Yes, we need a card (if we don't have one) for putting the expanded information on the Rover page into the CMS.

PR #1582, opened 2019-10-02
Improve accessibility on Rover page

RoyEJohnson:
Fixes #355
I ran a Lighthouse audit for the page and addressed the issues it found.

### DIFF
--- a/src/app/pages/rover-by-openstax/banner/banner.html
+++ b/src/app/pages/rover-by-openstax/banner/banner.html
@@ -4,8 +4,7 @@
     <div class="boxed">
         <div class="two-columns">
             <div class="left-column">
-                <img class="for-desktop" src="{model.headerImage}">
-                <img class="for-mobile" src="{model.mobileHeaderImage}">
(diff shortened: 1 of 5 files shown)


src/app/pages/rover-by-openstax/banner/banner.html
```diff
@@ -4,8 +4,7 @@
     <div class="boxed">
         <div class="two-columns">
             <div class="left-column">
-                <img class="for-desktop" src="{model.headerImage}">
-                <img class="for-mobile" src="{model.mobileHeaderImage}">
```
mwharrison: Does this mean we are no longer using the mobile header image? Or just on this page?

RoyEJohnson: I actually have it using a local (front-end) image rather than what's in the CMS, because it's an SVG. I was using that image for both images, so there wasn't any point to having different image elements for different displays.

mwharrison: Ok - this is approved. If I should remove anything from the CMS, just make a card for it.

PR #768, opened 2017-02-27
Restyle social media buttons

RoyEJohnson:

PR #1627, opened 2019-12-03
Fix date format

RoyEJohnson:
Safari does not accept 'yyyy-mm-dd' as a Date initializer. Need to use slashes.
Issue #768


mwharrison:
@RoyEJohnson, will this work fine with slashes on other browsers?

RoyEJohnson:
Yes, slashes are officially supported. Dashes are unofficially supported in browsers other than Safari.

RoyEJohnson:
@mwharrison Yes, slashes are officially supported. Dashes are _unofficially_ supported in browsers other than Safari.

PR #1630, opened 2019-12-04
Hotfix give link is external

RoyEJohnson:
This is cherry-picked from the hotfix that went out.


Brooke-Fail:
Verified in Staging and Production.

PR #1671, opened 2020-02-14
Fix partner Type value list

RoyEJohnson:
Was going to be a hotfix, but is just going into the next release.
Needs to be coordinated with a change in Salesforce.


mwharrison:
This needs a post-deployment task to change the API value on production to the new value.

PR #983, opened 2017-11-06
Randomize order of partners listed on book details page

RoyEJohnson:

PR #1730, opened 2020-04-06
Add highlighted resources section to instructor tab

RoyEJohnson:
Issue #983


src/app/helpers/controller/cms-mixin.js
```diff
@@ -50,7 +46,7 @@ async function getUrlFor(slug) {
 
     // A little magic to handle the news slug
     if (slug === 'news') {
-        apiUrl = await newsPromise;
+        apiUrl = '/apps/cms/api/pages/openstax-news';
```
mwharrison, line 49: is it possible to use the API base from the settings file (I think you have this somewhere)? That way there is only one place to change the API path if it changes.. again.

src/app/pages/details/details.js
```diff
@@ -64,6 +64,7 @@ export default class Details extends BaseClass {
 
         super.init();
         this.bookTitle = 'Loading';
+        console.info('Getting slug from', bookTitle);
```
mwharrison, line 67: debug message?

src/app/helpers/controller/cms-mixin.js
RoyEJohnson: Good catch

src/app/pages/details/details.js
RoyEJohnson: Got rid of a couple more.

PR #1733, opened 2020-04-09
Fix: ToC for Rex books doesn't change

RoyEJohnson:
lodash memoize only keys by the first parameter!
Issue #1023


RoyEJohnson:
This was tested in the hotfix.

PR #1743, opened 2020-04-29
Show hardcover and paperback ISBNs

RoyEJohnson:
Issue #529


RoyEJohnson:
@nathanstitt This is a tiny change, but if you could look over the whole file and leave any recommendations, that could help me write better React.

PR #1745, opened 2020-04-30
eliminate 30x redirects due to missing trailing slash

jpslav:
We've done good work over time to eliminate 30x redirects for our content on osweb (the ones that are just redirecting to add a trailing slash), but there are still 18 of them remaining.  They slow our site down and cost us in terms of requests made through Cloudfront.  Would be great if we could eliminate the remaining ones.  

Here's a stab at a few of them.

* I don't know where to fix the one for https://openstax.org/apps/cms/api/sticky?format=json
* I also can't see where to fix the ones for the images e.g. https://openstax.org/apps/cms/api/images/921

Ideally, a load of the home page should show no openstax.org 30x's in the chrome inspector.

### DIFF
--- a/src/app/components/shell/footer/footer.js
+++ b/src/app/components/shell/footer/footer.js
@@ -18,7 +18,9 @@ class Footer extends CMSPageController {
         /* eslint arrow-parens: 0 */
         (async () => {
             try {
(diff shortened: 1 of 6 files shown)


RoyEJohnson:
I piggybacked a tiny change to fix the APUSH title image on the details page, which doesn't automatically get as big as it can.

Issue #1765, opened 2020-06-02
Port About, Adopters, and Bookstore pages to React


RoyEJohnson:
# Description
These are simple pages, just starting at the top and working my way down.

# Acceptance Criteria
- Pages work as before (regression testing only)

brittanyearly:
Verified in QA.

brittanyearly:
Includes Contact Us form (and submission), Adoption form, Interest form, and their submissions. Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1766, opened 2020-06-03
Port about, adopters, and bookstore-suppliers pages to react

RoyEJohnson:
Issue #1765


src/app/helpers/controller/cms-mixin.js
```diff
@@ -99,6 +100,35 @@ export async function fetchPageData({slug, preserveWrapping, setsPageTitleAndDes
     return data;
 }
 
+export function fetchPageDataJsx(fpdParams) {
```
TomWoodward, line 103: because this function composes hooks, it is a custom hook, and react should be yelling at you about how it doesn't start with the word `use`

src/app/pages/confirmation/confirmation.js
```diff
@@ -23,70 +24,157 @@ const models = {
...
+    if (!errataId) {
+        return null;
     }
 
+    const email = fetchDefaultEmail(errataId);
+    const notifyByEmail = email && email !== 'none@openstax.org';
+
+    return (
+        notifyByEmail ?
+            `We'll be sending submission updates to ${email}.` :
```
TomWoodward, line 102: do you need some kind of loading state here? email will be undefined until the user model loads

src/app/pages/contact/form.jsx
```diff
@@ -0,0 +1,100 @@
...
+        routerBus.emit('navigate', '/confirmation/contact');
+    }
+
+    return (
+        <SalesforceForm postTo={postTo} afterSubmit={afterSubmit}>
+            <input type="hidden" name="external" value="1" />
+            <label>
+                What is your question about?
+                <Select>
+                    <select name="subject" ref={selectRef}>
```
TomWoodward, line 71: why does the `Select` component not render the `select` element?

src/app/helpers/controller/cms-mixin.js
RoyEJohnson: Interesting. I wonder if the lack of yelling is due to my using Preact?
I've corrected this one. Will have to be on the lookout for others.

src/app/pages/confirmation/confirmation.js
RoyEJohnson: If email is undefined, the user sees a link. If email successfully loads, that gets replaced with a message.

src/app/pages/contact/form.jsx
RoyEJohnson: Why not, indeed? Fixed.

src/app/helpers/controller/cms-mixin.js
RoyEJohnson: I've added the eslint plugin for React hooks to the next PR. That got me the yelling I was missing.

Issue #1782, opened 2020-06-25
Port blog page to React


RoyEJohnson:
# Description
Part of the ongoing port from Superb to React.
The Blog page is fairly complex. It should be a fair-sized project by itself.

# Acceptance Criteria
All parts of the Blog page are implemented in React

# Acceptance Tests
Regression only

brittanyearly:
Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1803, opened 2020-07-01
Port blog page to react

RoyEJohnson:
Issue #1782


src/app/pages/blog/article-summary/article-summary.jsx
```diff
@@ -0,0 +1,73 @@
...
+        const handleScroll = debounce(() => {
+            if (elRef.current && $.overlapsViewport(elRef.current)) {
+                window.removeEventListener('scroll', handleScroll);
+                setDelayedImage(image);
+            }
+        }, 200);
+
+        window.addEventListener('scroll', handleScroll);
+        handleScroll();
+    }, [image]);
```
nathanstitt, line 68: You'll need to return a function that will remove the event listener on unmount

this is also perfect candidate to make a custom `usesScrollPosition` hook.   https://dev.to/n8tb1t/tracking-scroll-position-with-react-hooks-3bbj is a decent walkthrough on that.  

Then for a more advanced one you could make the hook only add a single listener that would notify all the images when it's scrolled.  It might not be worth it to do that now though, but would be a good optimization

src/app/pages/blog/article/article.js
```diff
@@ -1,130 +1,159 @@
...
-        });
-        bylineRegion.attach(new Byline({
-            date: this.pageData.date,
-            author: this.pageData.author
-        }));
-        this.attachProgressRing();
-    }
+    return Math.round(100 * visibleHeight / totalHeight);
+}
+
```
nathanstitt, line 31: this could also be a hook that could use the proposed `useScrollPosition` internally.  Really it already is a hook because they're just functions, but a hook typically starts with `use`

src/app/pages/blog/article/article.js
```diff
@@ -1,130 +1,159 @@
...
+        handleScroll();
+
+        return () => {
+            window.removeEventListener('scroll', handleScroll);
+            window.removeEventListener('resize', handleScroll);
+        };
+    }, [bodyRef]);
+
+    return [progress, setBodyRef];
+}
```
nathanstitt, line 107: hm, looks like you already have a scroll hook?  in that case I guess maybe use this?

src/app/components/progress-ring/progress-ring.js
```diff
@@ -1,35 +1,37 @@
...
+                    fill="transparent"
+                    r={normalizedRadius}
+                    stroke-dasharray={`${circumference} ${circumference}`}
+                    stroke-width={stroke}
+                    style={`stroke-dashoffset: ${strokeDashoffset}`}
+                />
+            </svg>
+        </div>
+    );
 }
```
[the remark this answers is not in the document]
RoyEJohnson: Attributes should be kebab-case. When passed as a dictionary in JS, they are camelCase.

src/app/pages/blog/article-summary/article-summary.jsx
RoyEJohnson: Since I'm looking at the scroll position of the window object, I implemented a Context for window information (currently only scrollY, but in theory more things could be put in it). That gives me one scroll listener that notifies everybody. I'd appreciate your doing a sanity check on it.

src/app/pages/blog/article/article.js
RoyEJohnson: This one is just the helper function that figures out how far along in the article the page is. It is used by the hook.

src/app/pages/blog/article/article.js
RoyEJohnson: I made this use the WindowContext. I also fixed an awkward misunderstanding about forward refs (around `bodyRef`).

src/app/components/jsx-helpers/jsx-helpers.jsx
```diff
@@ -1,4 +1,34 @@
...
+}
+
+export const WindowContext = React.createContext(getValuesFromWindow());
+
+export function WindowContextProvider({children}) {
+    const [value, setValue] = useState(getValuesFromWindow());
+
+    useLayoutEffect(() => {
+        const handleScroll = () => setValue(getValuesFromWindow());
+
```
nathanstitt, line 16: the removeListeners looks great, but I think you'll also want to wrap the `setValue` function with a `debounce` so it isn't called too often.

PR #1809, opened 2020-07-17
Port press page to react

RoyEJohnson:


src/app/pages/press/article/article.js
```diff
@@ -1,52 +1,65 @@
...
-            title: this.pageData.title
-        };
-        this.update();
-        this.attachByline(this.pageData.author, formatDate(this.pageData.date));
-        this.pageData.body.forEach((bodyUnit) => {
-            this.regions.body.append(bodyUnitView(bodyUnit));
-        });
+    if (data && data.error) {
+        routerBus.emit('navigate', '/404', {path: '/blog'}, true);
+        return null;
```
nathanstitt, line 54: do you know that it's a 404 anytime there's an error?  Seemed strange to me that it's not checking anything about the error, but maybe that's the only time it occurs

src/app/pages/press/bookings/bookings.js
```diff
@@ -1,26 +1,55 @@
...
+        }));
+
+    return (
+        <div className="bookings">
+            <div className="heading">
+                <h2>Book our experts</h2>
+                <RawHTML html={blurb} />
+            </div>
+            {
+                bios.map((bio) => <Booking {...bio} />)
```
nathanstitt, line 51: also missing a `key`

src/app/pages/press/bookings/bookings.scss
```diff
@@ -7,4 +7,33 @@
...
     }
+
+    .booking {
+        display: grid;
+        grid-gap: $normal-margin;
+        grid-template-columns: auto 1fr;
+
+        .card {
+            @extend %card;
+
```
nathanstitt, line 18: @extend can create a lot of rules, you might make it a mixin to prevent that

src/app/pages/press/inquiries/inquiries.js
```diff
@@ -1,16 +1,59 @@
...
+        <div className="inquiries">
+            <div>
+                <h2>Press inquiries</h2>
+                <div className="contact">
+                    {
+                        Boolean(name) && <div>{name}</div>
+                    }
+                    <div>{phone}</div>
+                    <div>{email}</div>
+                </div>
```
nathanstitt, line 34: should you make phone and email be a link with the scheme of `tel:` and `email:`  ?

src/app/pages/press/mobile-selector/mobile-selector.js
```diff
@@ -1,45 +1,55 @@
...
-            selectedClass: (v) => v === this.props.selectedValue ? 'selected' : null
-        };
-    }
+function Menu({selectedValue, values, onChange}) {
+    return (
+        <div className="fixed-overlay">
+            <div role="menu">
+                {
+                    values.map((value) =>
+                        <MenuItem selectedValue={selectedValue} value={value} onChange={onChange}/>
```
nathanstitt, line 24: map needs `key` 😁 I also sometimes will use this type of prop if the variable names match properties:
` <MenuItem {...{ selectedValue, onChange, value }} />`

src/app/pages/press/press-excerpt/press-excerpt.js
```diff
@@ -1,42 +1,41 @@
...
-export default class PressExcerpt extends componentType(spec) {
+export default function PressExcerpt({iconUrl, author, date, source, url, headline, excerpt}) {
+    const ref = useRef();
 
-    onLoaded() {
-        if (this.iconUrl) {
-            this.el.classList.add('with-icon');
+    useEffect(() => {
+        if (iconUrl) {
+            ref.current.classList.add('with-icon');
```
nathanstitt, line 21: React is likely to not like you mucking with the DOM that it renders.  It seems like you could conditionally set `with-icon` in the class on the element when it renders instead.  the `classnames` lib makes that easy if you're using it

src/app/pages/press/article/article.js
RoyEJohnson: That's the only error I have available to me. Effectively, if I can't get the data, it's not found.

src/app/pages/press/inquiries/inquiries.js
RoyEJohnson: Good idea!

src/app/pages/press/press-excerpt/press-excerpt.js
RoyEJohnson: Yes, this was a wrong decision for how to modify class list. But I thought part of the reason `useEffect` existed was to say "I'm done with the DOM, you can fiddle with it now."

nathanstitt: as a general rule you shouldn't change anything on the DOM that react has rendered directly.  You'd probably be ok modifying just the class though, but if you made more substantial changes it's likely that react would have errors if it re-rendered the content and it didn't match what it expected it to look like.

As I understand it, useEffect means react has rendered the content so you're able to *read* from it by measuring and kicking off animations.

Issue #1789, opened 2020-06-25
Port Interest and Adoption forms to React


RoyEJohnson:
# Description
Part of the ongoing port from Superb to React

# Acceptance Criteria
All parts of the /interest and /adoption pages are implemented in React

# Acceptance Tests
Regression only (do ensure that the forms work as expected)

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1818, opened 2020-07-29
Adoption and Interest forms ported to React

RoyEJohnson:
Issue #1789, which was the year of the French Revolution


src/app/pages/interest/interest.js
```diff
@@ -1,178 +1,189 @@
...
+    const formRef = useRef();
+
+    function validatePage(page) {
+        const validateContactInfo = contactValidatorRef.current;
+
+        if (page === 1 && !validateContactInfo()) {
+            return false;
+        }
+        return true;
     }
```
nathanstitt, line 103: a quibble, but you could: `return Boolean(page !== 1 && validateContactInfo())`

src/app/pages/interest/interest.js
```diff
@@ -1,178 +1,189 @@
...
-            return contactForm.checkSchoolName() || contactForm.validate();
-        };
-        return result;
+        }
+
+        formRef.current = form;
+        if (iframe) {
+            iframe.addEventListener('load', submitAfterDelay);
+        }
+        submitNextBook();
```
nathanstitt, line 135: There's the possibility of a bug here if the form is unmounted in the 300ms after `submitAfterDelay` is called. I think you can work around that by checking if one of the references are still valid in `submitNextBook`.  I think that references are reset to null when unmounting, but not entirely sure about that.  

To test you could increase the timeout to a large number like 10 seconds, submit and then hit browser back button, then watch the console to see what happens.

src/app/components/form-input/form-input.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+        return () => shellBus.emit('no-sticky');
+    }, []);
+
+    return (
+        <div className="suggestions">
+            <div className="suggestion-box">
+                {
+                    !exactMatch && !accepted && matches.map((match, i) =>
+                        <SuggestionItem value={match} accept={accept} index={i}
```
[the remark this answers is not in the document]
RoyEJohnson: Clearly something I need to pay more attention to.

src/app/components/jsx-helpers/jsx-helpers.jsx
```diff
@@ -13,7 +14,7 @@ export function WindowContextProvider({children}) {
     const [value, setValue] = useState(getValuesFromWindow());
 
     useLayoutEffect(() => {
-        const handleScroll = throttle(() => setValue(getValuesFromWindow()), 40);
+        const handleScroll = () => setValue(getValuesFromWindow());
```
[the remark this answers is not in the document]
RoyEJohnson: No, this is not a good situation.

src/app/pages/interest/interest.js
RoyEJohnson: with ||, yes.

src/app/pages/interest/interest.js
RoyEJohnson: It looks like `selectedBooks` got cleared when the form unmounted, so the queue became empty. No complaints in the Console, at least.

PR #1828, opened 2020-08-06
Give all users the ability to request faculty access

BryanEliDimas:
**All** users, including previously rejected users, should have a way to get faculty access by clicking on a link which shows up in the login-menu. The link will say "Finish signing up" if faculty users have not even finished signing up (and therefore have not finished applying for faculty access) or "Request instructor access" for everyone else.

### DIFF
--- a/src/app/components/shell/header/main-menu/login-menu/login-menu.js
+++ b/src/app/components/shell/header/main-menu/login-menu/login-menu.js
@@ -62,10 +62,7 @@ class LoginMenu extends componentType(spec, busMixin) {
 
     attachLoginDropdown() {
         const facultySignupStep4 = `${settings.accountHref}/i/signup/educator/profile_form`;
-        const facultyAccessLink = `${settings.accountHref}/i/signup/educator/cs_form`;
-        const facultyAccessLinkLegacy = `${facultyAccessLink}/?is_legacy=true`;
-        const facultyAccessLinkStale = `${facultyAccessLink}/?is_stale=true`;
-        const rejectedFacultySignupStep4 = `${facultyAccessLink}/?prev_rejected=true`;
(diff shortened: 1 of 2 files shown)


src/app/components/shell/header/main-menu/login-menu/login-menu.js
```diff
@@ -72,13 +73,22 @@ class LoginMenu extends componentType(spec, busMixin) {
...
+            },
             {
                 label: 'Request instructor access',
-                url: facultyAccessLink,
+                url: reqFacultyAccessLink,
                 exclude: () => Boolean(
                     (this.user.groups || []).includes('Faculty') ||
-                    (this.user.groups || []).includes('Student') ||
-                    this.user.pending_verification
+                    (!this.user.stale_verification && this.user.pending_verification)
```
BryanEliDimas, line 91: Note: If my logic is right, this should show up for users who are "stale". @RoyEJohnson

BryanEliDimas: But at the same time, hide it for users who have not finished signing up

RoyEJohnson: It will show up if they're stale _or_ not pending. Is that right? Or should there be a ! on the pending as well, so it would show up if they're stale or pending?

BryanEliDimas: Correct, it should show up if they're stale or _not_ pending. Because if they're just pending and not stale, I don't want it to show up; in that case, that means they're in the middle of the signup process and thus I want the "Finish signing up" link to show up instead. 

(The default state from the beginning is pending, and then as they progress through the flow, it will be set to approved or rejected. Unless they are stale after 4 days stuck in pending.)

PR #1833, opened 2020-08-11
Microdonation messaging and links

RoyEJohnson:
Issue #1355


src/index.html
```diff
@@ -77,9 +77,11 @@
...
             </div>
```
nathanstitt, line 78: FYI, I've used https://github.com/gregberge/svgr to convert SVG's to react components before, it's a neat tool

RoyEJohnson: Thanks, will check it out.

PR #1863, opened 2020-09-25
Handle long titles in Adoption/Interest forms

RoyEJohnson:
This is a simple fix. It doesn't match the design, but it's better than before.


RoyEJohnson:
@mwharrison I've got a standing question about this for Victoria, as I couldn't match her design, but I think this is a definite improvement, and should go forward until she is available to give further refinements.

mwharrison (approved):

PR #1880, opened 2020-10-19
Remove 7-session limit for give banner

RoyEJohnson:


src/app/components/shell/lower-sticky-note/lower-sticky-note.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import {pageWrapper} from '~/controllers/jsx-wrapper';
 import {RawHTML} from '~/components/jsx-helpers/jsx-helpers.jsx';
-import {usePutAway, useStickyData, useSeenCounter} from '../shared.jsx';
+import {usePutAway, useStickyData} from '../shared.jsx';
 import analytics from '~/helpers/analytics';
 import './lower-sticky-note.css';
 
-const SEEN_ENOUGH = 7;
```
mwharrison, line 8: if this keeps getting added/taken away, maybe we could add something to the CMS?

mwharrison (approved):

PR #1890, opened 2020-10-28
Rc 2.69.0

RoyEJohnson:
Merge fixes back to master.

### DIFF
--- a/src/app/pages/details/common/hooks.js
+++ b/src/app/pages/details/common/hooks.js
@@ -35,6 +35,18 @@ function getUserStatusPromise() {
     };
 
     return userModel.load().then((user) => {
+        if (!user || Reflect.ownKeys(user).length === 0) {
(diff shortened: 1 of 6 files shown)


src/app/pages/details/common/hooks.js
```diff
@@ -35,6 +35,18 @@ function getUserStatusPromise() {
     };
 
     return userModel.load().then((user) => {
+        if (!user || Reflect.ownKeys(user).length === 0) {
```
mwharrison, line 38: do we need these debug messages?

mwharrison (approved):
one question - otherwise, approved

src/app/pages/details/common/hooks.js
RoyEJohnson: Yes, there's an elusive bug that Brittany has run into. The debug messages will give some clues if she runs into it.

PR #1908, opened 2020-12-07
Tutor marketing page redesign

RoyEJohnson:


src/app/pages/openstax-tutor/features/features.scss
```diff
@@ -0,0 +1,30 @@
+@import 'pattern-library/core/pattern-library/headers';
+
+.openstax-tutor-page .features {
+    display: flex;
+    grid-column-gap: 2rem;
```
nathanstitt, line 5: since this has been present on a flex element twice maybe it's intentional 😁

src/app/pages/openstax-tutor/materials/materials.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default function Materials({model: {availableBooksHeader: headline, tutorBooks: books}}) {
+    const headerLabel = 'Materials';
+
+    return (
+        <LabeledSection headerLabel={headerLabel} headline={headline}>
+            <div className="materials">
+                {
+                    books.map((book) =>
+                        <div className="book-card" key={book}>
```
nathanstitt, line 14: not sure about preact, but react requires keys to be strings. book is an object here

nathanstitt (approved):

src/app/pages/openstax-tutor/cost/cost.scss
```diff
@@ -0,0 +1,61 @@
+@import 'pattern-library/core/pattern-library/headers';
+
+.openstax-tutor-page .cost {
+    display: flex;
+    flex-direction: column;
+    grid-row-gap: 2rem;
```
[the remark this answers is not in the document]
RoyEJohnson: It does, though. I discovered that recently. It is documented: https://developer.mozilla.org/en-US/docs/Web/CSS/gap#Formal_definition

src/app/pages/openstax-tutor/materials/materials.js
RoyEJohnson: In Preact, any value can be used, though non-object-references are preferred. I should really use index for these, as they don't re-render. Really, I ought to turn off key for the line as a sort of pragma, I guess (which functionally uses index as key, without explicitly having to do so).
https://stackoverflow.com/a/42775667/392102

PR #1914, opened 2020-12-16
Add bri_book to login link when on LLPH page

RoyEJohnson:
Issue #1617

### DIFF
--- a/src/app/pages/llph/llph.js
+++ b/src/app/pages/llph/llph.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import {pageWrapper} from '~/controllers/jsx-wrapper';
 import {RawHTML, LoaderPage} from '~/components/jsx-helpers/jsx-helpers.jsx';
 import linkHelper from '~/helpers/link';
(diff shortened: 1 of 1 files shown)


src/app/pages/llph/llph.js
```diff
@@ -25,14 +25,24 @@ function LLPH({data}) {
 
     loginLocation.search = `?r=${encodeURIComponent(signupLinkHref)}&bri_book=LLPH`;
 
+    // Caution: spooky action at a distance
+    // Update login menu link
+    useEffect(() => {
```
RoyEJohnson, line 30: @nathanstitt I have a React element modify an unrelated non-React element. I don't like that, but I didn't see a better way. Curious whether anything suggests itself to you?

nathanstitt: this is where react's context (https://reactjs.org/docs/context.html) is handy.  You find a common parent, add a `menuLoginLinkHref` state,  and wrap it's children in a `Context.Povider` that exposes an object that contains current menu href as well as the `setState` call to update it.

That way the menu can read the state, and this component could set it.

nathanstitt (approved):

PR #1963, opened 2021-02-08
Use updated pattern library

RoyEJohnson:


src/app/pages/partners/partner-details/reviews/user-review.js
```diff
@@ -25,12 +25,13 @@ function UserControls({status}) {
...
 
+    // Temporarily disabling delete with "false" below.
     return (
         <React.Fragment>
             {status !== 'Approved' && <span className="review-status">{displayStatus}</span>}
             <div className="user-controls">
                 {
-                    status !== 'Deleted' &&
+                    false && status !== 'Deleted' &&
                         <React.Fragment>
```
nathanstitt, line 35: was the `false`  leftover from debugging?  Otherwise this will never render

nathanstitt (approved):
@RoyEJohnson I'll approve since the rest looks good, but I think the `false` might have been left in by mistake

src/app/pages/partners/partner-details/reviews/user-review.js
RoyEJohnson: Backed this commit out. Was a temporary thing that we have already reversed. Thought rebase would have taken it out. Thanks for catching it.

Issue #1943, opened 2021-01-22
Port FAQ page to React


RoyEJohnson:
# Description
The FAQ page and probably some others got missed earlier. Find unported pages and get them ported.

# Acceptance Criteria
Regression only

brittanyearly:
Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1971, opened 2021-02-24
Port FAQ page to React

RoyEJohnson:
Issue #1943


src/app/components/jsx-helpers/jsx-helpers.jsx
```diff
@@ -159,12 +159,13 @@ export function RawHTML({Tag='div', html, embed=false, ...otherProps}) {
 
 export function useToggle(initialState) {
     const [value, setValue] = useState(initialState);
+    const toggleRef = useRef(
+        (newValue=!value) => {
+            setValue(newValue);
+        }
+    );
```
nathanstitt, line 166: it looks like you're using useRef to remember the function so it's not re-created on each render.  For that  you typically use either `useMemo` or `useCallback` https://reactjs.org/docs/hooks-reference.html#usememo

nathanstitt (approved):
looks good, but think about replacing useRef since I think it'd confuse some people since its not actually remembering a dom element

src/app/components/jsx-helpers/jsx-helpers.jsx
RoyEJohnson: It doesn't seem to work using useCallback because the function needs to take an argument. useRef docs explicitly say they are useful for any mutable value. I'll keep my eyes open for proper uses of useCallback and useMemo.

Issue #1959, opened 2021-02-05
Port "general" pages to React


RoyEJohnson:
# Description
Part of the ongoing port to React

# Acceptance Criteria
General pages are implemented entirely in React

# Acceptance Tests
Regression only. The Tech Scout review standards is a General page (it isn't on Dev, though.)
We also have /general/test-general-page

brittanyearly:
Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1973, opened 2021-02-25
Port General Pages to React

RoyEJohnson:
Issue #1959

### DIFF
--- a/src/app/pages/general/general.js
+++ b/src/app/pages/general/general.js
@@ -1,45 +1,50 @@
-import componentType, {canonicalLinkMixin} from '~/helpers/controller/init-mixin';
-import css from './general.css';
-import $ from '~/helpers/$';
+import {pageWrapper} from '~/controllers/jsx-wrapper';
+import React from 'react';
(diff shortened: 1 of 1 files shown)


src/app/pages/general/general.js
```diff
@@ -1,45 +1,50 @@
...
+            el.appendChild(strips);
+        }
+        arr.push(el.outerHTML);
+        return arr;
+    }, []).join('\n');
+
+    return (
+        <RawHTML className="general page" html={innerHTML} embed />
+    );
+}
```
nathanstitt, line 27: hm, I haven't used `DOMParser` before, this is pretty neat.  Do you fully trust the html though? if not this looks like it could be a big security risk since it allows rendering arbitrary html

nathanstitt (approved):

src/app/pages/general/general.js
RoyEJohnson: This is for HTML pulled from the server, so yes. I would have just used RawHTML, but I had to identify and embellish the .block-heading piece.

Issue #1960, opened 2021-02-05
Port Institutional Partnership Application page to React


RoyEJohnson:
# Description
Part of the ongoing port to React

# Acceptance Criteria
- [x] All parts of the Institutional Parternship Application are implemented in React
- [x] delete tab-group.js (use only tab-group.jsx)
- [x] delete content-group.js (use only content-group.jsx)

# Acceptance Tests
Regression only. URL is /institutional-partnership-application

brittanyearly:
Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #1984, opened 2021-03-09
Port Institutional Partnership Application page to React

RoyEJohnson:
Issue #1960


src/app/pages/institutional-partnership-application/institutional-partnership-application.js
```diff
@@ -1,82 +1,118 @@
...
+    );
+}
+
+function Application() {
+    return (
+        <div className="application">
+            <iframe
+                class="form-iframe"
+                src="https://docs.google.com/forms/d/e/1FAIpQLSfI0XpCC-4ag0u9vHizT5w2JLkwcv39HVTZnvQHEtmQj8PJwQ/viewform"
+            />
```
nathanstitt, line 51: huh, I didn't know you could iframe google docs!

src/app/pages/institutional-partnership-application/institutional-partnership-application.js
```diff
@@ -1,82 +1,118 @@
...
+        </div>
+    );
+}
 
-export default class Institutional extends BaseClass {
+function ProgramDetails({model}) {
+    return (
+        model[0].map((section) =>
+            <div key={section.heading}>
```
nathanstitt, line 59: will the headings always be unique?  Something I've learned with react is that having duplicate keys can cause really weird re-rendering.  Luckily it'll complain about that as well as missing ones, but not sure what preact does

nathanstitt (approved):

src/app/pages/institutional-partnership-application/institutional-partnership-application.js
RoyEJohnson: I get red error messages in the console for duplicate keys. In this case, headings really should be unique.

PR #1987, opened 2021-03-16
Port creator fest page to react

RoyEJohnson:
And throw out a lot of no-longer-used Superb code.


src/app/pages/creator-fest/fetched-content/fetched-content.js
```diff
@@ -1,70 +1,34 @@
...
-    css,
-    get view() {
-        return {
-            tag: 'section'
-        };
-    }
-};
+// March 16 2021: Removing Signup form code. This page is currently inactive with
+// no way to see whether porting worked. If and when we re-activate the page,
+// pull code from older archive
```
nathanstitt, line 8: nice comment! I was wondering where all the code went

src/app/pages/creator-fest/home-content/home-content.js
```diff
@@ -1,43 +1,17 @@
...
-            if (i === 0) {
-                Object.assign(model, {registerBox});
-                SectionConstructor = About;
+export default function HomeContent({pagePanels, register}) {
+    return (
+        <div className="home">
+            <About data={pagePanels[0]} register={register} />
+            {
+                pagePanels.slice(1).map((data) =>
+                    <Why data={data} key={data.superheading} />
```
nathanstitt, line 12: same question as on one of the other pro, `superheating` seems like it might not be unique, maybe it'd be better to just use the index of the element here

nathanstitt (approved):

src/app/pages/creator-fest/home-content/home-content.js
RoyEJohnson: Sounds good to me.

PR #1992, opened 2021-03-23
Fix ratings synopsis for partners with no ratings

RoyEJohnson:
Issue #1782


nathanstitt (approved):
I've got a rule that I try hard to follow, whenever I make a bug fix I also add a test to show it's fixed, there's nothing worse than having a bug resurface later.  This is simple enough that maybe it doesn't warrant that though

PR #1994, opened 2021-03-24
Add Sentry

RoyEJohnson:


src/app/sentry.js
```diff
@@ -0,0 +1,9 @@
+import * as Sentry from '@sentry/react';
+import {Integrations} from '@sentry/tracing';
+
+Sentry.init({
+    dsn: 'https://8468b55bf5b2442d9d31840113399b10@o484761.ingest.sentry.io/5691260',
```
nathanstitt, line 5: should this also be read from env?

nathanstitt (approved):

src/app/sentry.js
RoyEJohnson: I don't *think* so, but this was my first experience with Sentry.

mwharrison: @RoyEJohnson it would be good to put this in the env - although it's not super secret, it can still be used nefariously.

mwharrison: note: you can generate a new key easily (and should) here: https://sentry.io/settings/openstax/projects/osweb/keys/

RoyEJohnson: @mwharrison Is pu something I need Scott to set up? I create the key, have him put it into the environment?

mwharrison: @RoyEJohnson I'm not sure how the FE reads secrets from the env - do you have anything else that does this? @nathanstitt how is tutor doing this?

PR #2009, opened 2021-04-19
Throttle login checks

RoyEJohnson:
Issue #1814

### DIFF
--- a/src/app/models/usermodel.js
+++ b/src/app/models/usermodel.js
@@ -1,6 +1,7 @@
 import {useState, useEffect} from 'react';
 import bus from './usermodel-bus';
 import isEqual from 'lodash/isEqual';
+import throttle from 'lodash/throttle';
 
 const settings = window.SETTINGS;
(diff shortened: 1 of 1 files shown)


src/app/models/usermodel.js
```diff
@@ -134,7 +135,9 @@ window.addEventListener('focus', () => {
             }
         });
     });
-});
+}, 90000);
```
nathanstitt, line 138: is that 90 seconds?  That seems like a pretty long amount of time to wait for accounts, but I'm probably missing something

nathanstitt (approved):

src/app/models/usermodel.js
RoyEJohnson: The check happens just in case you went to another tab and logged in. If you leave the tab and return very often, you can rack up a bunch of checks. 90 seconds could be a bit long, but it's an uncommon case, and most of the time you won't have just left, returned, then left and logged in.

PR #2025, opened 2021-05-20
Add Ally information/application page

RoyEJohnson:


src/app/helpers/controller/cms-mixin.js
```diff
@@ -126,3 +126,20 @@ export function usePageData(fpdParams) {
...
+    const [text, setText] = React.useState();
+
+    React.useEffect(() => {
+        const url = urlFromSlug(slug);
+
+        fetch(url).then((r) => r.text()).then((pageHtml) => {
+            const parser = new window.DOMParser();
+            const newDoc = parser.parseFromString(pageHtml, 'text/html');
+
+            setText(newDoc.body.innerHTML);
```
nathanstitt, line 140: I'm curious what the advantage of using DomParser vs the old-school method of blindly setting innerHTML to the string?  Maybe you want it to raise the exception if parsing fails?

nathanstitt (approved):

src/app/helpers/controller/cms-mixin.js
RoyEJohnson: In this case, I'm pulling an entire page, head and all, and just want to use the body.

PR #2042, opened 2021-06-09
Add donation ask to pdf click

RoyEJohnson:
Issue #1844

### DIFF
--- a/src/app/components/dialog/dialog.js
+++ b/src/app/components/dialog/dialog.js
@@ -75,33 +75,36 @@ export default function Dialog({
 
 export function useDialog() {
     const [showDialog, updateShowDialog] = React.useState(false);
-    const open = () => updateShowDialog(true);
-    const close = () => updateShowDialog(false);
 
-    function BoundDialog({
-        title, children, modal = true, className,
-        showPutAway=true,
-        ...otherProps
-    }) {
-        const Modal = modal ? ReactModal : React.Fragment;
+    return React.useMemo(() => {
+        const open = () => updateShowDialog(true);
+        const close = () => updateShowDialog(false);
 
-        return (
-            <Modal
-                isOpen={showDialog}
-                className='modal'
-                overlayClassName="modal-overlay"
-                bodyOpenClassName="no-scroll-dialog"
-                {...otherProps}
-            >
-                <Dialog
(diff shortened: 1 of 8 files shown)


src/app/components/dialog/dialog.js
```diff
@@ -71,3 +72,36 @@ export default function Dialog({
...
+            >
+                <Dialog
+                    title={title} className={className} isOpen={showDialog}
+                    onPutAway={showPutAway && close}
+                >
+                    {children}
+                </Dialog>
+            </Modal>
+        );
+    }
```
nathanstitt, line 104: you'll get a new instance of `BoundDialog`, open & close on every render, this will force react to unmount & remount on every render since the component will change. If the returned `BoundDialog` didn't change it would just re-render and probably not need to change any DOM.  I think you could use useMemo here to implement that

src/app/pages/details/common/get-this-title-files/give-before-pdf/thank-you-form.js
```diff
@@ -0,0 +1,115 @@
...
+
+function ThankYou({userModel={}, link, close}) {
+    const first = userModel.first_name;
+    const last = userModel.last_name;
+    const school = userModel.self_reported_school;
+
+    function afterSubmit() {
+        window.open(link);
+        close();
+    }
```
nathanstitt, line 52: sneaky stuff here.  Are you sure the the window.open will be actually open, since it's opening inside an invisible iframe?  Sure seems like that's exactly what content blockers will try to prevent

nathanstitt (approved):
look good, a bit wondering about the window.open but guessing you've tested that and it does work ok

src/app/pages/details/common/get-this-title-files/give-before-pdf/thank-you-form.js
RoyEJohnson: The `window` object is still the global context. It doesn't change to refer to the iframe just because an iframe event is calling the function. What's going on is that the form submission response loads in the iframe; when that happens, then the new window opens displaying the PDF.

src/app/components/dialog/dialog.js
RoyEJohnson: I will revisit this.

PR #2051, opened 2021-07-06
Add my openstax page

RoyEJohnson:


src/app/pages/my-openstax/main-card/navigator/navigator.js
```diff
@@ -0,0 +1,49 @@
...
+    const elPoll = window.setInterval(() => {
+        if (el) {
+            window.clearInterval(elPoll);
+        } else {
+            setEl(document.getElementById(id));
+        }
+    }, 40);
+
+    return el;
+}
```
nathanstitt, line 20: I think should this might need:
  * To be enclosed in a useEffect that returns a fn to clear the interval on unmount
  * have a timeout setting to set an upper bound on how long it'll keep retrying.  That way if the element is never found it'll stop at some point

src/app/pages/my-openstax/store/sfapi.js
```diff
@@ -0,0 +1,49 @@
+const server = 'https://dev.salesforce.openstax.org'; // 'http://localhost:4004'
+//  /api/v1/opportunities/0060v000008WUp1AAG
```
nathanstitt, line 2: just curious, is there a local mock server for salesforce?  if so, that might be useful for tutor

nathanstitt (approved):
looks good but check the loop in the useElementById

src/app/pages/my-openstax/main-card/main-card.scss
```diff
@@ -0,0 +1,98 @@
...
+  .layout-grid {
+    @include width-up-to($tablet-max) {
+      grid-template: 'banner'
+      'nav'
+      'main';
+    }
+
+    @include wider-than($tablet-max) {
+      grid-template: 'banner banner' auto
+      'nav main' auto / auto 1fr;
```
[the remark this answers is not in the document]
RoyEJohnson: That's the division operator; this one divides row specs from column specs.

PR #2062, opened 2021-07-21
Context loader for blog and details pages

RoyEJohnson:


src/app/pages/blog/blog-context.js
```diff
@@ -73,12 +73,4 @@ function useLocationSynchronizedToPath() {
...
 
-export function BlogContextProvider({children}) {
-    const value = useLocationSynchronizedToPath();
-
-    return (
-        <BlogContext.Provider value={value} children={children} />
-    );
-}
-
 export default BlogContext;
```
nathanstitt, line 76: I've found it's convenient to export a `useBlogContext` hook so that consumers only need a single import and can't muck it up

src/app/components/jsx-helpers/context-loader.js
```diff
@@ -0,0 +1,26 @@
...
+
+export default function ContextLoader({Context, slug, useContextValue, children, ...loaderArgs}) {
+    const Child = ({data}) => (
+        <ContextProvider {...{Context, data, useContextValue, children}} />
+    );
+
+    return (
+        <LoaderPage slug={slug} Child={Child} {...loaderArgs} />
+    );
+}
```
nathanstitt, line 26: I just wrote a little loader for labs that you might find interesting.  It has a built-in delay so that it only displays after a given timeout.  That's helpful for preventing the case where the API returns very quickly and you get a flash of a loader before the content snaps in.  

https://github.com/openstax/labs/blob/main/frontend/src/components/loading-animation.tsx#L185-L195 and https://github.com/openstax/labs/blob/main/frontend/src/lib/hooks.ts#L3

nathanstitt (approved):

src/app/pages/blog/blog-context.js
RoyEJohnson: Currently, they `import BlogContext` and call `React.useContext(BlogContext)`. Is your idea to wrap the `useContext` in its own hook?

nathanstitt: yah that's what I often do.  It just seems cleaner to me to have a single `useBlogContext()` call vs `React.useContext(BlogContext)`    With the later you have to import from both the BlogContext and useContext. not a big deal either way of course

PR #2069, opened 2021-07-30
Add language features

RoyEJohnson:


src/app/components/hero/hero.js
```diff
@@ -1,39 +1,39 @@
...
+//     const wrapped = innerWidth <= wrapNarrowerThan;
+//
+//     React.useLayoutEffect(() => {
+//         if (el && !wrapped && isTooSkinny(el)) {
+//             setWrapNarrowerThan(innerWidth);
+//         }
+//     }, [el, wrapped, innerWidth]);
+//
+//     return wrapped;
+// }
```
nathanstitt, line 30: bunch of commented code that could probably be removed?

src/app/components/jsx-helpers/page-loader.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {fetchFromCMS} from '~/helpers/controller/cms-mixin';
+import $ from '~/helpers/$';
+
+export default function usePageData(slug, preserveWrapping=false) {
+    const [data, setData] = React.useState();
+
+    React.useEffect(() => {
+        fetchFromCMS(slug, preserveWrapping).then($.camelCaseKeys).then(setData);
+    }, [slug, preserveWrapping]);
```
nathanstitt, line 10: you might consider adding a `debounce` to the `fetchFromCMS` function.  If the props change rapidly you wouldn't want to fire 1,000's of network requests

src/app/pages/details/desktop-view/desktop-view.js
```diff
@@ -5,23 +5,41 @@ import DetailsTab from './details-tab/details-tab';
...
+        'Student resources',
+        'Videos'
+    ],
+    es: [
+        'Detalles del libro',
+        'Recursos del instructor',
+        'Recursos para estudiantes',
+        'Videos'
+    ]
+};
```
nathanstitt, line 27: if this starts to become common, we should move all the strings into a json/yaml file where translators could work on them

nathanstitt (approved):
You should consider moving the transition strings to a json file and then import and use them from there.  Webpack supports importing a JSON file so you could do stuff like:
```js
import { QuickLink } from 'translations.json'

const {language} = useLanguageContext();
const labels = QuickLink[language];

```


where JSON looks like:
```json
{
"QuickLink": {
   "en": [ ... ],
   "es": [ ... ]
}
}
```

Or even better chat with REX to see what they've used for translations with React.  I'm pretty sure their solution has more features like fallbacks in case a translation is missing

src/app/components/jsx-helpers/page-loader.js
RoyEJohnson: This is for loading a page. I don't expect those props to change. But it does occur to me that I might use different tools to set this up so that it's more clear. 🤔

src/app/pages/details/desktop-view/desktop-view.js
RoyEJohnson: I peppered a lot of translations throughout the code as a stopgap. Ultimately, most of them should be added to the CMS and removed from the FE.

Issue #1771, opened 2020-06-16
Menu height/centering is wrong


RoyEJohnson:
The main menu is somewhat clipped so that the logo is not centered and the green highlights for the menu are too low.

victoria-cooke:
@RoyEJohnson Looks good

brittanyearly:
Verified in QA.

brittanyearly:
Verified in Staging.

brittanyearly:
Verified in Production.

PR #2079, opened 2021-08-09
Restyle cookie popup

RoyEJohnson:
Issue #1771

### DIFF
--- a/src/app/components/dialog/dialog.js
+++ b/src/app/components/dialog/dialog.js
@@ -17,7 +17,7 @@ function PutAway({noTitle, onClick}) {
 }
 
 export function FooterDialog({
-    isOpen, title, children
+    isOpen, title, children, className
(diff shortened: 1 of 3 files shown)


src/app/components/dialog/dialog.js
```diff
@@ -28,10 +29,13 @@ export function FooterDialog({
 
     footerEl.style.zIndex = 1;
     return (
-        <dialog className="footer-dialog">
-            <div className="title-bar">
-                <RawHTML Tag="span" html={title} />
-            </div>
+        <dialog className={classes}>
```
nathanstitt, line 32: not a big deal at all, but we typically call `cn` right in the prop, like:
`<dialog className={cn('footer-dialog', className)}>`

nathanstitt (approved):

PR #2080, opened 2021-08-09
Recognize invalid books

RoyEJohnson:
When loading book data fails, route to 404.
Issue #1886

### DIFF
--- a/src/app/components/shell/shell.js
+++ b/src/app/components/shell/shell.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import 'preact/debug';
 import {LanguageContextProvider} from '~/models/language-context';
 import {BrowserRouter} from 'react-router-dom';
 import Router from './router';
--- a/src/app/components/shell/shell.js
+++ b/src/app/components/shell/shell.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import 'preact/debug';
 import {LanguageContextProvider} from '~/models/language-context';
 import {BrowserRouter} from 'react-router-dom';
 import Router from './router';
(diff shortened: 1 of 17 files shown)


src/app/components/shell/shell.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import 'preact/debug';
```
nathanstitt, line 2: should this be left in the build for production?

src/app/helpers/controller/cms-mixin.js
```diff
@@ -98,31 +98,22 @@ export async function fetchPageData({slug, preserveWrapping, setsPageTitleAndDes
...
-            try {
-                setPageData(await fetchPageData(fpdParams));
-            } catch (err) {
-                setPageDataError(err);
-            }
-        }
-        fetchData();
+        fetchPageData(fpdParams).then(setPageData);
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [fpdParams.slug]);
```
nathanstitt, line 106: just in case you didn't know about it, there's also `eslint-disable-line`. I like using that for lines that are short because then it'll be deleted along with the line if it's removed:
`}, [fpdParams.slug]); // eslint-disable-line react-hooks/exhaustive-deps`

src/app/pages/blog/article/article.js
```diff
@@ -126,9 +126,10 @@ export function Article({data}) {
 
 function ArticleLoader({slug}) {
     const [data, statusPage] = usePageData({slug, preserveWrapping: true});
+    const {fail} = useRouterContext();
 
     if (data && data.error) {
-        routerBus.emit('navigate', '/404', {path: '/blog'}, true);
+        fail();
```
nathanstitt, line 132: using the `fail()` is a neat idea, might steal it

nathanstitt (approved):

PR #2091, opened 2021-08-19
Code cleanup

RoyEJohnson:
This is mostly about using buildContext to create contexts, rather than the more ad hoc way I did earlier.


nathanstitt (approved):
all looks ok to me.  🤞

PR #2092, opened 2021-08-23
Use more accessible tooltip for errata correction schedule

RoyEJohnson:
Need to pull data from CMS

### DIFF
--- a/src/app/pages/errata-summary/hero/hero.js
+++ b/src/app/pages/errata-summary/hero/hero.js
@@ -1,16 +1,11 @@
 import React, {useState, useEffect} from 'react';
-import {RawHTML} from '~/components/jsx-helpers/jsx-helpers.jsx';
-import cmsFetch from '~/models/cmsFetch';
+import {RawHTML, LoaderPage} from '~/components/jsx-helpers/jsx-helpers.jsx';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faInfoCircle} from '@fortawesome/free-solid-svg-icons/faInfoCircle';
 import bookPromise from '~/models/book-titles';
 import useRouterContext from '~/components/shell/router-context';
-
-const instructions = 'Errata submissions are displayed below until a new PDF is published online.';
-const moreAbout = 'More about our correction schedule';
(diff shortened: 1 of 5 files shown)


src/app/pages/errata-summary/hero/hero.js
```diff
@@ -34,29 +29,80 @@ function useBookInfo(book) {
     return info;
 }
 
-function useErrataHoverHtml(slug) {
-    const [html, setHtml] = useState('<p>...loading...</p>');
+const middle = '-135';
+
+function shiftIntoView(ref, leftOffset, setLeftOffset) {
+    const {left, right} = ref.current.getBoundingClientRect();
+    const {right: pageRight} = ref.current.closest('.page').getBoundingClientRect();
```
nathanstitt, line 36: what's your browser support like?  will it have `closest` ?  ie 11 doesn't have that

src/app/pages/errata-summary/hero/hero.js
```diff
@@ -34,29 +29,80 @@ function useBookInfo(book) {
...
+        <div className="poptip-attachment">
+            <div className="poptip" ref={ref} style={style} role="tooltip" aria-hidden={!isOpen}>
+                <div className="title-bar">
+                    More information
+                </div>
+                <RawHTML className="body" html={html} />
+            </div>
+        </div>
+    );
+}
```
nathanstitt, line 69: is this the wrapper element?  if so should it have `aria-alert` set?

nathanstitt (approved):

src/app/pages/errata-summary/hero/hero.js
RoyEJohnson: We don't support IE, only Edge. Good point, though.

src/app/pages/errata-summary/hero/hero.js
RoyEJohnson: This is not an alert. I have set role to `tooltip`.

nathanstitt: seems like role="tooltip" does the same though, so never mind!

PR #2094, opened 2021-08-26
Use router instead of native history and location

RoyEJohnson:
Use history and location features of the React Router rather than the native ones.


nathanstitt (approved):
Looks good to me Roy!

PR #2104, opened 2021-09-09
Fix book cover menus

RoyEJohnson:


src/app/pages/details/common/resource-box/resource-boxes.js
```diff
@@ -17,27 +17,41 @@ import useUserContext from '~/contexts/user';
...
+
     return (
         <div className="resource-box double">
             <div className="top-line">
-                <h3>{model.heading}</h3>
-                <img src={model.logoUrl} alt="" />
+                <h3>{heading}</h3>
+                <img src={logoUrl} alt="" />
             </div>
             <div className="blurb">
```
nathanstitt, line 41: I see you're setting one of the urls to null below, should this code test for null and skip showing the img in that case?  I'm not 100% sure what  does with a null url, hopefully it doesn't show an ugly broken image 😬

nathanstitt (approved):
looks good but might check the null thing

PR #2117, opened 2021-09-27
Send adoption and interest forms to pardot

RoyEJohnson:


src/app/components/contact-info/contact-info.js
```diff
@@ -1,43 +1,93 @@
...
+                        ]}
+                        selectedValue={schoolLocation}
+                        setSelectedValue={setSchoolLocation}
+                        required
+                    />
+                </div>
+            }
+        </React.Fragment>
+    );
+}
```
nathanstitt, line 85: it's a matter of preference but I dislike huge conditionally rendered blocks, because it's too hard to tell what gets rendered.  I find it cleaner to break each block out into a component and either pass a `isHidden` prop or have it contain the rendering logic internally.  Either way it can simply `return null` when it shouldn't render.

src/app/pages/adoption/adoption.js
```diff
@@ -75,49 +58,30 @@ function FacultyForm({selectedRole, onPageChange}) {
...
+                <React.Fragment>
+                    <input type="hidden" name="application_source" value="OS Web" />
+                    <input type="hidden" name="position" value={position} />
+                    <input type="hidden" name="role" value="Instructor" />
+                    <input type="hidden" name="lead_source" value="Adoption Form" />
+                    <input type="hidden" name="process_adoptions" value={true} />
+                    <ContactInfo validatorRef={contactValidatorRef} />
+                </React.Fragment>
+                <BookSelectorPage selectedBooksRef={selectedBooksRef} />
+            </MultiPageForm>
```
nathanstitt, line 83: would be nice to extend `MarkupChildren` to handle multiple children, that way you wouldn't need to use React.Fragment everywhere with `MultiPageForm`

src/app/pages/interest/interest.js
```diff
@@ -68,83 +75,51 @@ function BookSelectorPage({selectedBooksRef, bookBeingSubmitted}) {
             <FormInput
                 longLabel="How many students do you teach each semester?"
                 inputProps={{
-                    name: 'Number_of_Students__c',
+                    name: 'number_of_students',
                     type: 'number',
```
nathanstitt, line 79: just curious, but did SF get it's fields renamed so they don't use the double underscores now?

nathanstitt (approved):
looks good, just some comments

PR #2124, opened 2021-10-01
General page 404s

RoyEJohnson:


src/app/helpers/controller/cms-mixin.js
```diff
@@ -125,12 +125,17 @@ export function useTextFromSlug(slug) {
 
         fetch(url)
             .catch((err) => setText(err))
```
nathanstitt, line 127: should `catch` should come after the `.then`?  if it did then it'd also catch any JS errors that might occur inside the `then` block, like maybe if the html was invalid and parser failed?

nathanstitt (approved):

PR #2137, opened 2021-11-22
Limit adoptions (and interest) to 5 books

RoyEJohnson:


src/app/components/book-selector/book-selector.js
```diff
@@ -22,31 +23,53 @@ function Subject({subject, books, name, selectedBooks, toggleBook}) {
...
+                <div className="hint">{hintText(selectedBooks.length, limit)}</div>
             </div>
             {
                 subjects.map((subject) =>
                     <Subject
                         key={subject}
                         subject={subject} books={booksBySubject(subject)} name={name}
-                        selectedBooks={selectedBooks} toggleBook={toggleBook}
+                        selectedBooks={selectedBooks} toggleBook={checkLimitAndToggle}
                     />
```
nathanstitt, line 73: I see you're using a disallowed cursor style, but should `BookCheckbox` above also be disabled when limit is reached?

nathanstitt (approved):
a question about disabling the inputs, but looks good

src/app/components/book-selector/book-selector.js
RoyEJohnson: You're right, it should be a disabled state. I've updated the code.

PR #2142, opened 2021-12-03
Improve handling of invalid urls

RoyEJohnson:


src/app/components/shell/router.js
```diff
@@ -109,37 +109,38 @@ function useLinkHandler() {
...
+    }, [isRedirect]);
 }
 
-function useHomeOrMyOpenStax() {
-    const user = useMyOpenStaxUser();
+// Dynamic import is persnickety about building paths; you can't just pass it
+// a string with a path, it needs to know it's building a string
+// Then webpack chunks everything that matches that string (in this case, all pages)
+// in one chunk. Not as dynamic as one might hope.
+function useImportedComponent(name) {
```
nathanstitt, line 128: I'm not sure if this will work, but I've also fought with this and the solution I came up with was passing a function that calls import around instead of a string.  So this could look like:
```
function useImportedComponent(importFn) {
    ...
}
```

and then be called like:
```
const page =  ImportedPage(useMemo(() => import('~/pages/foo/bar')));
```

nathanstitt (approved):

src/app/components/shell/router.js
RoyEJohnson: I had something like that when I was trying to load the General page and the Normal page in parallel. That did not go well, so I wound up back with just one import function.

PR #2144, opened 2021-12-08
Sentry error fixes

RoyEJohnson:
Upgraded a bunch of packages, including Webpack.
Addressed all Sentry errors that have been reported (so far) in the latest release.


src/app/models/usermodel.js
```diff
@@ -68,14 +68,20 @@ const accountsModel = {
...
+                            }
+                            return result;
+                        },
+                        (err) => {
+                            console.warn('No JSON in Accounts response');
+                            return {err};
                         }
-                        return result;
-                    });
+                    );
```
nathanstitt, line 84: interesting, this is essentially what kinetic does as well.  I wonder how hard it would be to expose the user's json as a `window.user` global or something so kinetic and other pages JS could use it.  It's not clear if this is fetched on every page though

nathanstitt (approved):

src/app/models/usermodel.js
RoyEJohnson: I created a UserContext that wraps the main Router code. That calls this; almost all user-checking code uses the Context. I should check that the few things that still do use it directly need to.

PR #2146, opened 2021-12-09
Add cache-busting parameter for settings.js

RoyEJohnson:


mwharrison (approved):
I wonder if there is a way we could do this more automated? Maybe during the switch to the new deployment, we can append the version number or something.

PR #2148, opened 2021-12-10
add compression webpack plugin

RoyEJohnson:


RoyEJohnson:
The server doesn't seem to want to deal with this. May revisit some other time.

PR #2149, opened 2021-12-13
Use WEBP to replace PNG and JPG

RoyEJohnson:
And delete unused images


nathanstitt (approved):
Oh this is interesting, it looks like safari < 14 has slipped to only `0.78` % usage.

You might still consider adding the [polyfill](https://github.com/chase-moskal/webp-hero), but it's really heavy weight IMO

PR #2152, opened 2021-12-17
Refactor multi page form

RoyEJohnson:


src/app/components/multi-page-form/multi-page-form.js
```diff
@@ -1,138 +1,167 @@
...
+                        onPageChange
+                    }}
+                >
+                    <MarkupChildren children={children} />
+                    <PageCount />
+                    {
+                        submitting &&
+                            <div className="big-message">Submitting...</div>
+                    }
+                    <ButtonRow disabled={submitting} {...{formRef, onSubmit}} />
```
nathanstitt, line 154: BTW, I wrote a fancy "pendingState" hook for Kinetic that waits for 150ms to display.  the concept behind that is that you don't want a spinner to display for very short requests because it causes flickering.  Our [LoadingAnimation](https://github.com/openstax/kinetic/blob/main/frontend/src/components/loading-animation.tsx#L185) uses it internally.  Maybe it'll be of use to you

https://github.com/openstax/kinetic/blob/3d0c1de92383365a8fb5a699e2180dfcadd7c2be/frontend/src/lib/hooks.ts#L4-L19

nathanstitt (approved):

PR #2157, opened 2022-01-07
Merge hotfix back to main

RoyEJohnson:
Primarily resurrects the Creator Fest page


src/app/pages/creator-fest/creator-fest.js
```diff
@@ -1,10 +1,54 @@
...
+import HomeContent from './home-content/home-content';
 import './creator-fest.scss';
 
-export default function CFPlaceholder() {
+function PageContent({data, navLinks}) {
+    // Gives reactivity on window.location.pathname
+    useLocation();
+    const linkEntry = navLinks
+        .find((obj) => `/creator-fest/${obj.url}` === window.location.pathname);
+
```
nathanstitt, line 15: it might be a small improvement to have `useLocation` return `window.location` and then use the value here instead of reading `window.location`.  Doing so would be a bit clearer and would also set you up for eventual server-side rendering later

nathanstitt (approved):

src/app/pages/creator-fest/creator-fest.js
RoyEJohnson: Good catch. `useLocation` _does_ return `location`. So I'm using that instead of `window.location`.

PR #2159, opened 2022-01-11
Fix handling special characters in Subjects filter

RoyEJohnson:


src/app/pages/subjects/book-viewer/book-viewer.js
```diff
@@ -4,46 +4,46 @@ import {ActiveElementContextProvider} from '~/contexts/active-element';
...
+                        result[cmsCategory].push(book);
+                    }
+                });
             }
-        });
-    }
 
-    addLabels();
+            for (const category of categories) {
+                if (result[category.cms]) {
```
nathanstitt, line 29: above on line 19 you used `in` to test for the key, I assumed that was because the value might be 0/false so you needed to test if the key existed.  Probably should do the same here?

src/app/pages/subjects/book-viewer/book-viewer.js
```diff
@@ -4,46 +4,46 @@ import {ActiveElementContextProvider} from '~/contexts/active-element';
...
     const subjectHtml = categoryData.html;
     const books = categorizedBooks[categoryData.cms] || [];
 
-    if (!['view-all', categoryData.value].includes(category)) {
-        classList.push('hidden');
-    }
-
     return (
-        <div className={classList.join(' ')}>
+        <div className={cn('book-category', {hidden})}>
```
nathanstitt, line 46: a tiny thing, but I find it clearer to move the test as close to where it's used as possible.  maybe this would be clearer:
```jsx
<div 
  className={cn('book-category', { 
    hidden: !['view-all', categoryData.value].includes(category),
  })}
>
```

nathanstitt (approved):

src/app/pages/subjects/book-viewer/book-viewer.js
RoyEJohnson: I actually prefer the encapsulation on things like this.

src/app/pages/subjects/book-viewer/book-viewer.js
RoyEJohnson: To be completely correct, I need to test in and typeof.

PR #2163, opened 2022-01-13
Limit display of takeover banner

RoyEJohnson:


src/app/components/shell/takeover-dialog/takeover-dialog.js
```diff
@@ -9,8 +9,36 @@ import DesktopContent from './content-desktop';
 import MobileContent from './content-mobile';
 import './takeover-dialog.scss';
 
+const RECENT_DELTA_MS = 16 * 60 * 60 * 1000; // 16 hours
+const LS_KEY = 'takeoverLastDisplay';
+
+function useLocalStorage(key, defaultValue='') {
+    const storedValue = JSON.parse(window.localStorage.getItem(key));
```
nathanstitt, line 16: I've learned the hard way to always wrap all localStorage parsing in a `try/catch` block, with a sane fallback value.  In the wild I've seen it get corrupted and no longer contain valid json

nathanstitt (approved):
approved, but you probably should add try/catch :D

PR #2165, opened 2022-01-13
Scroll to top when new page loads

RoyEJohnson:
The trick is when a new page loads, not simply a pathname change.

### DIFF
--- a/src/app/components/shell/router.js
+++ b/src/app/components/shell/router.js
@@ -156,10 +156,18 @@ function FallbackToGeneralPage({name}) {
 
 function ImportedPage({name}) {
     const {Component, loadError} = useImportedComponent(name);
+    const history = useHistory();
 
     useAnalyticsPageView();
 
-    useEffect(() => window.scrollTo(0, 0), [name]);
+    useEffect(
+        () => {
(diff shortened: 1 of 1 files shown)


src/app/components/shell/router.js
```diff
@@ -159,6 +159,8 @@ function ImportedPage({name}) {
 
     useAnalyticsPageView();
 
+    useEffect(() => window.scrollTo(0, 0), [name]);
+
```
nathanstitt, line 163: I think doing this will also scroll to the top when you hit the back button.  In that case you probably want to leave the page scrolled to it's original position.  To do so you could use the stateObject parameter of push/replace state to store the scroll position, then restore it later.

That's assuming you have control over the push/replace State call, not sure if you do.

nathanstitt (approved):

src/app/components/shell/router.js
RoyEJohnson: Good thinking. React Router does have a solution. :)

RoyEJohnson: Or, actually, it doesn't. :(

PR #2181, opened 2022-02-18
Use Spanish Coming Soon badge

RoyEJohnson:


src/app/pages/details/common/resource-box/resource-box.scss
```diff
@@ -28,9 +28,8 @@
         }
     }
 
-    &.coming-soon::before {
-        @include coming-soon;
-
+    &.coming-soon::before,
+    &.coming-soon-es::before {
```
nathanstitt, line 32: hm, how I'd consider doing this is to have the language set on a class way up on document.body or on a high-level wrapper.  Then you could test for: ` @at-root .lang-es &.coming-soon` (I think that's how @at-root works)

that would remove the requirement to set the lang on every single component that's translated.

nathanstitt (approved):
looks ok, but you might consider setting language on root element

PR #2183, opened 2022-02-24
Fix webtolead debug mode (testimonial form)

RoyEJohnson:


RoyEJohnson:
We removed the testimonial form.

PR #2188, opened 2022-03-28
Changes to deploy frontend to S3

Dantemss:
- Update ruby to non-EOL version
- Use contenthash instead of fullhash (to prevent deprecation warning)
- Prefix frontend asset URLs with /dist/

I think the only change going forward is that when writing URLs to images and assets you'll have to use `/dist/images/` instead of `/images/`.


Dantemss:
I think the only change going forward is that when writing URLs to images and assets you'll have to use `/dist/images/` instead of `/images/`.

RoyEJohnson (approved):

.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.7.4
```
RoyEJohnson, line 1: I think we could actually remove Ruby from the dist entirely. It's not really used. It was just part of the standard setup stuff that Phil included.

mwharrison (approved):

Dantemss:
Due to the path changes, once this is merged we will need to use the separate FE deployment going forward (run the deploy-fe script once per environment to set it up, then update the main distribution, then use deploy-fe after that to update it).

PR #2195, opened 2022-04-12
Blog: PDF viewer, Video

RoyEJohnson:
Also most of Gated Content, but there is more to be done there.


nathanstitt (approved):
react-pdf seems like a really neat project!  I'm looking forward to seeing this deployed

PR #2198, opened 2022-04-18
Latest blog posts page

RoyEJohnson:


src/app/pages/blog/blog.js
```diff
@@ -21,13 +27,15 @@ export function SearchResultsPage() {
...
-    const {search} = useLocation();
 
-    if (search) {
-        return (<SearchResultsPage />);
-    }
+    useEffect(
+        () => {document.title = 'OpenStax News';},
+        []
+    );
+
```
nathanstitt, line 38: I was like 99% sure that Tutor had a DocumentTitle component it rendered for this, but I can't seem to find it 😭 

Still, might be nice to extract. is probably just something like:
`const DocumentTitle = ({ title ) =>  useEffect(() => document.title = title, [])`

nathanstitt (approved):

src/app/pages/blog/blog.js
RoyEJohnson: Interesting to make that a component.

PR #2210, opened 2022-05-05
Layout renewal form

RoyEJohnson:
Not including some back end data items


src/app/components/multiselect/book-tags/book-tags.scss
```diff
@@ -25,6 +25,7 @@
         border: thin solid text-color(link);
         border-radius: 0.3rem;
         display: flex;
+        gap: 1rem;
         min-height: 3rem;
```
nathanstitt, line 29: yay for flex gap!

nathanstitt (approved):

PR #2213, opened 2022-05-09
Reduce time to interactive

RoyEJohnson:


src/app/components/jsx-helpers/jsx-helpers.jsx
```diff
@@ -66,12 +66,37 @@ export function LoaderPage({
...
+        newScript.async = false;
+        s.parentNode?.replaceChild(newScript, s);
+
+        p.then(processOne);
+    });
+
+    processOne();
+};
+
 export function RawHTML({Tag='div', html, embed=false, ...otherProps}) {
```
nathanstitt, line 94: I've used https://www.npmjs.com/package/loadjs for things like this before

src/app/components/shell/welcome/welcome-content.js
```diff
@@ -0,0 +1,122 @@
...
+
+    function skipWalkthrough() {
+        cookie.setKey('walkthroughDone');
+        goToMyOpenStax();
+    }
+
+    function showWalkthrough() {
+        window.localStorage.showMyOpenStaxWalkthrough = true;
+        goToMyOpenStax();
+    }
```
nathanstitt, line 30: Keep in mind that safari won't allow localStorage in an incognito window, so don't do anything that depends on it.  I've also been using https://github.com/astoilkov/use-local-storage-state which is pretty nice

nathanstitt (approved):

src/app/components/jsx-helpers/jsx-helpers.jsx
RoyEJohnson: I think that's for a different purpose. This is to allow RawHTML with `<script>` pieces -- usually Wistia video players.

nathanstitt: hah, I thought you were inserting script tags to load js

PR #2223, opened 2022-06-03
Handle equity rating filter

RoyEJohnson:


src/app/components/accordion-group/accordion-group.js
```diff
@@ -29,7 +30,33 @@ function useChevronDirection(forwardOnChange, preExpanded) {
     return [chevronDirection, onChange];
 }
 
-function TitleBar({title, titleTag, chevronDirection}) {
+function TooltipButton({content}) {
+    const ref = React.useRef();
+    const focus = React.useCallback(
+        (e) => {
+            ref.current.focus();
+            e.stopPropagation();
```
nathanstitt, line 38: do you ever envision a use-case where the button is inside an `a` tag or somewhere that a click should bubble up to?   If so, you probably don't want `stopPropagation()`

nathanstitt (approved):
looks good but maybe a bug in the results.js, it's not new though so I'm probably reading the code wrong

PR #2231, opened 2022-06-16
Clear up labeling for reviews and ratings

RoyEJohnson:


src/app/components/stars-and-count/stars-and-count.js
```diff
@@ -54,7 +54,7 @@ export default function StarsAndCount({rating, count=0, showNumber=false}) {
         <div className="stars-and-count">
             <Stars stars={rating} />
             {showNumber && (count > 0) && roundedRating(rating)}
-            <span>({count || 'no reviews'})</span>
+            <span>({`${count || 'no'} reviews`})</span>
```
nathanstitt, line 57: This will read a bit strange when there's `1 reviews` Does your fancy new i18n framework have a `pluralize` function?  Otherwise I've used https://www.npmjs.com/package/plur before.

nathanstitt (approved):

PR #2249, opened 2022-07-22
refactor out storeon

RoyEJohnson:


src/app/models/usermodel.js
```diff
@@ -25,46 +25,46 @@ function cached(fn) {
...
+                }
+            ],
+            applications: [
+                {
+                    id: 5,
+                    name: 'OpenStax CMS Dev'
+                }
+            ]
+        });
         // eslint-disable-next-line no-unreachable
```
nathanstitt, line 68: heh, I'm guessing this was left enabled by mistake 😁

nathanstitt (approved):

nathanstitt:
approved pretty sure you'll need to re-comment out the stubbed user

src/app/models/usermodel.js
RoyEJohnson: I knew there would come a day that I would wind up committing that. Thanks.

PR #2253, opened 2022-07-26
Fix Kinetic page tab title and description

RoyEJohnson:
This applies to all general pages


nathanstitt (approved):
I'm sorry that our rename broke this, I wasn't aware of it 😭

PR #2255, opened 2022-08-01
Increase Partners page height so Books popover doesn't overlap footer

RoyEJohnson:
In addition to the layout fix,
A bug in the release highlighted a lack of test coverage in the book publication info.
This PR covers that test and some other additional testing.


nathanstitt (approved):
LGTM!

PR #2257, opened 2022-08-04
Add blogs and webinars to specific-subject page

RoyEJohnson:


src/app/components/carousel/carousel.js
```diff
@@ -59,9 +59,31 @@ function Carousel({
...
+                (i) => {
+                    const {right} = i.getBoundingClientRect();
+
+                    return right - left <= vpWidth;
+                }
+            );
+
+            return Math.max(Number(atATime), wholeItems.length);
+        }
+    )();
```
nathanstitt, line 87: I'm not sure, but using IntersectionObserver might be a good candidate here.  You could add it has a callback that's triggered whenever an element appears inside the bounding div.

Its also unclear to me why you're using an  Immediate-Invoked Function Expression here?  Why not assign to `step` directly?

nathanstitt (approved):

src/app/components/carousel/carousel.js
RoyEJohnson: The immediate-invoked function came from converting it from useMemo. I need to extract it to a separate function and call it.
I don't think IntersectionObserver buys me anything I need; I just need to know how many items are in the viewport so I can scroll by that many items when they click. Thinking about it now, I should _probably_ put that in the frame-changer code, but I wrote it to be entirely unaware of the container and items.

PR #2263, opened 2022-08-11
Blog page explore sections and pages

RoyEJohnson:


src/app/pages/blog/blog-context.js
```diff
@@ -1,15 +1,94 @@
...
+
+function useTopicStories() {
+    const [topicType, setTopicType] = React.useState();
+    const [topic, setTopic] = React.useState();
+    const setTypeAndTopic = React.useCallback(
+        (typ, top) => {
+            setTopicType(typ);
+            setTopic(top);
+        },
+        []
```
nathanstitt, line 45: I think `setTopic` and `setTopicType` should be listed in the dependencies array here, right?  I'm surprised that the linter didn't complain .

src/app/pages/blog/explore-page/explore-page.js
```diff
@@ -0,0 +1,97 @@
...
+                </HeadingAndSearchBar>
+                <div className="explore-topic-blurb text-content">
+                    {subject?.pageContent}
+                </div>
+                <PinnedArticle subhead={heading} />
+                <div className="popular-posts">
+                    <SectionHeader head="Popular blog posts" subhead={heading} />
+                    <div className="latest-blurbs cards">
+                        {
+                            topicPopular.map(blurbModel).map((article) =>
```
nathanstitt, line 85: instead of double `map`, you could combine in one by wrapping the return value of `blurbModel`:

```jsx
topicPopular.map((a) => (
const article = blurbModel(a) && (
 <div className="card" key={article.articleSlug}>
   <ArticleSummary {...{...article, setPath, HeadTag: 'h3'}} />
</div>
))
```

Of course would be much cleaner to extract the summary into it's own component and do that inside it. then this would just be:
```
<div className="latest-blurbs cards">{ topicPopular.map(article => <Article article={article} />)}</div>
```

.. hehe, now after writing this all out, maybe the double map isn't so bad 🤷

nathanstitt (approved):

PR #2265, opened 2022-08-18
Use instructor resource API with hidden-links feature

RoyEJohnson:
Check for Resource tab/pane display when window width changes


nathanstitt (approved):
@RoyEJohnson sorry I let this take so long, somehow I'd missed the request

PR #2275, opened 2022-09-12
Decruft code causing Sentry errors

RoyEJohnson:


nathanstitt (approved):
love the term "decruft" 🤣

src/app/components/shell/shared.jsx
```diff
@@ -78,7 +78,7 @@ function useCampaign(stickyData) {
     const mode = getMode(stickyData);
 
     useEffect(() => {
-        if (mode) {
+        if (mode && window.localStorage) {
```
nathanstitt, line 81: good old safari private browsing mode?

PR #2277, opened 2022-09-14
Rework callout counter

RoyEJohnson:


src/app/pages/separatemap/map-api.js
```diff
@@ -0,0 +1,146 @@
+import mapboxgl from 'mapbox-gl';
+import mapboxPromise from '~/models/mapbox';
+// import useMapContext from './map-context';
```
chrisbendel, line 3: Could remove this?

chrisbendel (approved):

nathanstitt (approved):
Looks good,  a few general questions though:
 * what's the rationale behind using React.<thing> vs importing the function and using it?  It seems like you've been transitioning over the last few pr?  
 * Beware that using useMemo doesn't always speed up anything and may make things slower.  A re-render may be faster than the memoization : https://blog.logrocket.com/rethinking-hooks-memoization/

nathanstitt:
Looks good,  a few general questions though:
 * what's the rationale behind using React.<thing> vs importing the function and using it?  It seems like you've been transitioning over the last few pr?  
 * Beware that using useMemo doesn't always speed up anything and may make things slower.  A re-render may be faster than the memoization : https://blog.logrocket.com/rethinking-hooks-memoization/

RoyEJohnson:
> @nathanstitt 
It's purely preference. In a lot of cases, I wind up using, for example, useEffect once, and it's actually less typing to just add the six characters "React." than ", {useEffect}". But more generally, I don't like to have to make a change in two places if I can just make it in one.

I appreciate the article. I think I'm doing my memoization _mostly_ right. I am aware of the fact that primitives generally don't need to be memoized. I might have some more things to look out for.

PR #2279, opened 2022-09-20
Rework partners data

RoyEJohnson:


src/app/pages/partners/active-filters/filter-remover.js
```diff
@@ -4,7 +4,7 @@ export default function FilterRemover({label, store, value}) {
     return (
         <div className="filter-remover">
             <span>{label}</span>
-            <button type="button" onClick={() => store.toggle(value)}>
+            <button type="button" aria-label="remove" onClick={() => store.toggle(value)}>
```
nathanstitt, line 7: I've just discovered there's a new `data-action` attribute for things like this, you could add `data-action="delete"`: https://www.w3.org/TR/adapt-content/#action-explanation.

src/app/pages/partners/analytics-events.js
```diff
@@ -1,52 +1,81 @@
...
+        [books]
+    );
+    const SAERef = React.useRef();
 
-function sendAddEvent(actionObj) {
-    if ('add' in actionObj) {
-        sendFilterEvent(actionObj.add);
-    }
-}
+    SAERef.current = React.useCallback(
```
nathanstitt, line 22: should it cause a re-render when `SAERef` is set?  if so you'll need to store it in state, ref assignment doesn't cause a re-render

src/app/pages/partners/partner-details/partner-details.js
```diff
@@ -4,7 +4,7 @@ import Synopsis from './synopsis/synopsis';
 import Carousel from './carousel/carousel';
 import Reviews from './reviews/reviews';
 import {RawHTML} from '~/components/jsx-helpers/jsx-helpers.jsx';
-import TabGroup from '~/components/tab-group/tab-group.jsx';
+import TabGroup from '~/components/tab-group/tab-group';
```
nathanstitt, line 7: you can set a lint rule to catch this: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md

src/app/pages/partners/partner-details/reviews/user-review.js
```diff
@@ -13,19 +13,23 @@ function UserControls({status}) {
...
-    }
-
-    function onDelete(event) {
-        event.preventDefault();
-        // eslint-disable-next-line no-alert
-        if (window.confirm('Are you sure you want to delete this review?')) {
-            postRating({id: myReview.id}, 'DELETE');
-        }
-    }
+    const onUpdate = React.useCallback(
```
nathanstitt, line 16: FYI, tutor had a nifty `SuretyGuard` component that overlays a button with an "are you sure" popup: https://github.com/openstax/tutor-js/blob/main/shared/src/components/surety-guard.js

I'll probably steal it in kinetic when/if we have a need for it

nathanstitt (approved):
Looks good Roy!

src/app/pages/partners/analytics-events.js
RoyEJohnson: I'm using SAERef to avoid having `sendAddEvent` get re-called each time the `sendFilterEvent` callback updates.

PR #2292, opened 2022-10-04
Fix hamburger menu

RoyEJohnson:


src/app/components/shell/header/menus/main-menu/dropdown/dropdown.js
```diff
@@ -1,15 +1,20 @@
...
 import RawHTML from '~/components/jsx-helpers/raw-html';
 import useDropdownContext from '../../dropdown-context';
+import useWindowContext from '~/contexts/window';
 import {isMobileDisplay} from '~/helpers/device';
+import {treatSpaceOrEnterAsClick} from '~/helpers/events';
 import cn from 'classnames';
 import './dropdown.scss';
 
 export function MenuItem({label, url, local}) {
+    const {innerWidth: _} = useWindowContext();
```
chrisbendel, line 11: is useWindowContext doing something behind the scenes that we need? looks like innerwidth is unused

chrisbendel (approved):

src/app/components/shell/header/menus/main-menu/dropdown/dropdown.js
RoyEJohnson: It is unused (which is why I name it "_", a permissible unused variable name). It signals that the window size has updated, so the window size-checker (which is not itself reactive) will do its check.

PR #2296, opened 2022-10-14
Merge changes from 10 November release to main

RoyEJohnson:


src/app/helpers/page-data-utils.js
```diff
@@ -99,7 +99,7 @@ export function useTextFromSlug(slug) {
 
                         setText(newDoc.body.innerHTML);
                         setHead({
-                            title: newDoc.head.querySelector('title').textContent,
+                            title: newDoc.head.querySelector('title')?.textContent,
```
chrisbendel, line 102: should this default to anything in case the queryselector doesnt return anything?

chrisbendel (approved):

src/app/helpers/page-data-utils.js
RoyEJohnson: There is a default when the title value is used.

PR #2325, opened 2022-12-19
migrate research page to os-webview repo

chrisbendel:


RoyEJohnson (commented):
Did you consider using WEBP for the portraits? Also, do you plan to put them into the CMS?

chrisbendel:
@RoyEJohnson sorry - forgot to mention, all these images will be removed from the repo! I'll clean that up on the next commit

src/app/pages/research/components/header.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import {colors, media} from '~/pages/research/theme';
+import styled from '@emotion/styled';
+import {Box} from 'boxible';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {Section} from '~/pages/research/components/section';
+import {faArrowUpRightFromSquare} from '@fortawesome/free-solid-svg-icons';
```
RoyEJohnson, line 7: `import {faArrowUpRightFromSquare} from '@fortawesome/free-solid-svg-icons/faArrowUpRightFromSquare';
`

src/app/pages/research/components/members.js
```diff
@@ -0,0 +1,258 @@
...
+import {faChevronUp} from '@fortawesome/free-solid-svg-icons/faChevronUp';
+import {faChevronDown} from '@fortawesome/free-solid-svg-icons/faChevronDown';
+import Dialog from '~/components/dialog/dialog';
+import {isMobileDisplay} from '~/helpers/device';
+
+const MemberGrid = styled.div({
+    display: 'grid',
+    gap: '2rem',
+    padding: '4rem',
+    gridTemplateColumns: 'repeat(auto-fit, minmax(140px, 1fr))',
```
RoyEJohnson, line 17: Consider using `rem` instead of `px`

src/app/pages/research/components/members.js
```diff
@@ -0,0 +1,258 @@
...
+        <Box direction='column' align='center' className='text-center'>
+            <img
+                height={145}
+                width={145}
+                alt={member.photo.title}
+                src={member.photo.file}
+                tabIndex={0}
+                onClick={() => {
+                    setShow(true);
+                    return false;
```
RoyEJohnson, line 140: You should probably be using `event.preventDefault()` rather than returning false.
https://deepscan.io/docs/rules/react-bad-event-handler-return-false/

RoyEJohnson (changes requested):
Did you consider using WEBP for the portraits? Also, do you plan to put them into the CMS?

src/app/pages/learning-research/learning-research.js
```diff
@@ -0,0 +1,46 @@
...
+import {ResearchSection} from '~/pages/learning-research/components/research-areas';
+import {Funders} from '~/pages/learning-research/components/funders';
+import {ContactUs} from '~/pages/learning-research/components/contact-us';
+import {Publications} from '~/pages/learning-research/components/publications';
+import {MembersSection} from '~/pages/learning-research/components/members';
+
+function LearningResearchPage({data}) {
+    return (
+        <React.Fragment>
+            <LazyLoad>
```
RoyEJohnson, line 15: Generally speaking, there's no reason to lazy load the first element of a page. Only the elements that are not in the viewport when the page loads benefit.

src/app/pages/learning-research/research.scss
```diff
@@ -0,0 +1,106 @@
+@import 'pattern-library/core/pattern-library/headers';
+
+@include wider-than($phone-max) {
+    .mobile-only {
+        display: none;
+    }
+
+    .ReactModal__Content {
```
RoyEJohnson, line 8: I have some concerns that after loading, these `!important` specs will override values for other pages, since they're not scoped to a particular className here.

RoyEJohnson (changes requested):
I think the dialog styling needs to be associated with a className so it isn't a global change.

src/app/pages/learning-research/components/funders.js
```diff
@@ -0,0 +1,55 @@
...
+const SupporterImage = styled.img({
+    width: '100%',
+    height: 'auto'
+});
+
+export const Funders = () => {
+    return (
+        <Section>
+            <Box direction="column" align={{ mobile: 'center' }} className='py-2 container'>
+                <div
```
nathanstitt, line 21: same as above, we can remove py-2 and use padding prop

src/app/pages/learning-research/components/header.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const Banner = ({data: {bannerBody, bannerHeader, bannerCTA, bannerURL}}) => {
+    if (!bannerBody) {return null;}
+    return (
+        <div css={{backgroundColor: colors.lightTeal}}>
+            <Box direction={{mobile: 'column'}} className='container align-items-center py-2' gap='medium'>
+                <h4 css={{color: colors.blackText, flex: 1}}>
+                    {bannerHeader}
+                </h4>
+                <Box align={{mobile: 'center'}} className='justify-content-center' direction='column' css={{flex: 4}}>
```
nathanstitt, line 43: the `flex` property accepts a string, so this could be `flex="4"`

src/app/pages/learning-research/components/header.js
```diff
@@ -0,0 +1,89 @@
...
+                    </a>}
+                </Box>
+            </Box>
+            <OXColoredStripe />
+        </div>
+    );
+};
+
+export const OXColoredStripe = () => (
+    <Box css={{ height: 10 }}>
```
nathanstitt, line 62: Box has a `height` property that defaults to px

src/app/pages/learning-research/components/header.js
```diff
@@ -0,0 +1,89 @@
...
+                backgroundColor: colors.yellow,
+                flex: 2.5
+            }}></span>
+        <span
+            css={{
+                backgroundColor: colors.lightBlue,
+                flex: 1.5
+            }}></span>
+    </Box>
+);
```
nathanstitt, line 89: @RoyEJohnson do you have the OpenStax colored bar implemented somewhere already?

src/app/pages/learning-research/components/members.js
```diff
@@ -0,0 +1,252 @@
...
+};
+
+const MemberImage = styled.img({
+    marginBottom: '.5rem',
+    width: 145,
+    height: 145,
+    [media.mobile]: {
+        width: 75,
+        height: 75,
+        marginRight: '2rem'
```
nathanstitt, line 182: instead of using marginRight to space, should it use flexbox `gap` property?  That also has the advantage of spacing vertically if the images wrap

src/app/pages/learning-research/components/publications.js
```diff
@@ -0,0 +1,79 @@
...
+                </Box>
+            </a>
+            {publication.github &&
+                <a className='text-decoration-none' href={publication.github} target='_blank' rel="noreferrer">
+                    <Box align='center'>
+                        Github&nbsp;
+                        <FontAwesomeIcon size='sm' icon={faArrowUpRightFromSquare} />
+                    </Box>
+                </a>}
+        </Box>
```
nathanstitt, line 77: using a run-on non breaking space makes the words a bit hard to read here.  Maybe it'd be better to wrap the text in a span and set `gap` on the wrapping Box.

```jsx
<Box align='center' gap>
  <span>Github</span>
  <FontAwseomeIcon ... />
</Box>
```

src/app/pages/learning-research/components/research-areas.js
```diff
@@ -0,0 +1,68 @@
...
+        </Section>
+    );
+};
+
+export const ResearchFocusAreas = ({researchAreaList}) => {
+    return (
+        <TabAccordionCombo collapseAll={true}>
+            {researchAreaList.map((researchArea, i) => {
+                return (
+                    <div label={researchArea.researchAreaTitle} selected className='py-2' key={i}>
```
nathanstitt, line 24: I've don't think a `label` attribute is valid on a div tag?  Maybe something in the website's JS code's reading it though?

webpack.config.js
```diff
@@ -78,50 +78,50 @@ const config = {
...
+            context: (p) => p.startsWith('/cms/assets/fonts'),
+            target: API_ORIGIN,
+            changeOrigin: true
+        },
+        static: {
+            directory: path.resolve(__dirname, './src'),
+            publicPath
+        }
+    }
+};
```
nathanstitt, line 125: @RoyEJohnson do these lint changes seem correct to you?  it seems like it some of the removals might be undesired, such as the trailing commas.

src/app/pages/learning-research/components/research-areas.js
chrisbendel: label is used in the `TabAccordionCombo` as a prop

src/app/pages/learning-research/components/header.js
chrisbendel: found it, there is an image that we can use instead of this :)

webpack.config.js
RoyEJohnson: I don't generally use trailing commas in osweb. Run `yarn eslint` and `yarn stylelint` to ensure everything is still happy. The build system doesn't always error out on style errors.

RoyEJohnson (approved):

PR #2336, opened 2023-01-18
Change end year to current year

RoyEJohnson:


RoyEJohnson:
I made a change in the FE to find the second year (the digits after a dash) and replace them with the current year.

chrisbendel (approved):

PR #2354, opened 2023-02-24
Add K12 menu item

RoyEJohnson:
Infrastructure changes:
Code formatting rules are being removed from linters. The responsibility for code formatting (where spaces go, what goes on a single line, etc.) now falls to a separate utility. In this case, prettier.
Also swapped esbuild-loader for babel-loader, speeding up build times a bit.


RoyEJohnson:
@chrisbendel For this review, you should probably just look at the first commit. The second commit is just the result of running the code formatter.

chrisbendel (approved):

PR #2362, opened 2023-03-22
First pass assignable

RoyEJohnson:


src/app/pages/assignable/assignable.scss
```diff
@@ -0,0 +1,56 @@
...
+            margin: 0;
+        }
+    }
+
+    .btn.primary {
+        @extend %button;
+        @extend %primary;
+    }
+
+    .near-white {
```
chrisbendel, line 33: just a nitpick but are these colors available in some more global-esque css file?

src/app/pages/assignable/sections/overlapping-quote/overlapping-quote.scss
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        .name {
+            font-weight: bold;
+        }
+    }
+
+    .big-orange-quote {
+        @include set-font(hero-h1);
+
```
chrisbendel, line 64:
```suggestion
```

chrisbendel (approved):

src/app/pages/assignable/assignable.scss
RoyEJohnson: Reasonable point. We have something that's a little less white, which I think improves contrast. I'll use that.

PR #2371, opened 2023-04-13
Add react loadable

RoyEJohnson:
No significant functional changes. Just refactoring.


src/app/components/shell/router-helpers/fallback-to-general.js
```diff
@@ -0,0 +1,16 @@
...
+export default function FallbackToGeneralPage({name}) {
+    const [fallback, setFallback] = useToggle(false);
+
+    if (fallback) {
+        return <Error404 />;
+    }
+
+    return (
+        <GeneralPageFromSlug slug={`spike/${name}`} fallback={setFallback} />
+    );
```
bethshook, line 15: some of these components might read a little more concisely if you used ternaries rather than if statements, but that's just a personal preference

bethshook (approved):

PR #2381, opened 2023-05-03
Add some analytics metadata

TomWoodward:
I'm working on adding shared analytics-events code via GTM and [this other helper](https://github.com/openstax/analytics-helpers). meaning all we have to do in osweb is add the metadata as data attributes.

there is a cms change (https://github.com/openstax/openstax-cms/pull/1446) to add the cnx_id to the default book data response, the analytics won't work until both changes are in place but nothing breaks without the supporting change.

### DIFF
--- a/src/app/helpers/tag-manager.js
+++ b/src/app/helpers/tag-manager.js
@@ -22,23 +22,7 @@ const tagManagerID = 'GTM-W6N7PB';
     }
 })(window, document, 'script', 'dataLayer', tagManagerID);
 
-// GA4 stuff
-const {GA4: ga4id} = window.SETTINGS;
-
-if (ga4id) {
-    const tag = document.createElement('script');
-
-    tag.type = 'text/javascript';
-    tag.onload = () => {
-        window.dataLayer = window.dataLayer || [];
-
-        function gtag(...args) {
-            window.dataLayer.push(args);
-        }
-
-        gtag('js', new Date());
-        gtag('config', ga4id);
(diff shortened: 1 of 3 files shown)


src/app/helpers/tag-manager.js
```diff
@@ -22,23 +22,7 @@ const tagManagerID = 'GTM-W6N7PB';
...
-
-        gtag('js', new Date());
-        gtag('config', ga4id);
-    };
-    tag.src = `https://www.googletagmanager.com/gtag/js?id=${ga4id}`;
-    document.body.appendChild(tag);
-}
+window.dataLayer = window.dataLayer || [];
+// eslint-disable-next-line prefer-rest-params
+window.gtag = window.gtag || function () {window.dataLayer.push(arguments);};
```
RoyEJohnson, line 27: Why not use the rest params (`...args`)?

TomWoodward: there is some magic going on here, because calling `dataLayer.push` directly doesn't behave the same way as calling gtag. so i figured we should do it as close the [the actual instructions](https://developers.google.com/tag-platform/gtagjs/install) as possible. i'm actually more scared about the function not having a name, but it seems to work this way and i was having some problem i forget with doing it the other way

RoyEJohnson (approved):

RoyEJohnson (approved):

RoyEJohnson (approved):

PR #2386, opened 2023-05-09
"No upcoming webinars" card

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2241

### DIFF
--- a/src/app/pages/webinars/webinars.js
+++ b/src/app/pages/webinars/webinars.js
@@ -28,8 +28,9 @@ function Webinars({data: {heading: headline, description, heroImage}}) {
     const index = firstFuture > 0 ? firstFuture : webinarData.length;
     const upcomingData = webinarData.slice(index);
     const pastData = webinarData.slice(0, index);
+    const emptyMessage = webinarData.noWebinarsMessage || '(No webinars)';
     const tabContents = [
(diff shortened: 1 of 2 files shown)


RoyEJohnson:
This is written to work with a CMS value that don't exist yet, but fall back to a default value.

src/app/pages/webinars/webinars.js
```diff
@@ -28,8 +28,9 @@ function Webinars({data: {heading: headline, description, heroImage}}) {
     const index = firstFuture > 0 ? firstFuture : webinarData.length;
     const upcomingData = webinarData.slice(index);
     const pastData = webinarData.slice(0, index);
+    const emptyMessage = webinarData.noWebinarsMessage || '(No webinars)';
```
jivey, line 31: Am I looking at this right, the array will get this as a property? That feels unintuitive but maybe makes sense with how the JSON data is structured?

jivey: If the JSON is easy to modify, this might feel better as an object:
```javascript
{ emptyMessage: '(No webinars)', events: [...the current webinar data] }
```

RoyEJohnson: The empty message will be pulled separately, so I moved it closer to where it's used.

jivey (approved):

PR #2394, opened 2023-05-12
Fix locked resource box issue

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2238

The issue here is that the unlocked flag was not at the top level of the data, but within the `resource` member.

### DIFF
--- a/src/app/pages/details/common/resource-box/resource-box.js
+++ b/src/app/pages/details/common/resource-box/resource-box.js
@@ -45,7 +45,7 @@ function resourceBoxPermissions({
 // is available to the user (instructor version)
 export function instructorResourceBoxPermissions(resourceData, userStatus, search) {
     const resourceStatus = () => {
-        if (resourceData.resourceUnlocked || userStatus.isInstructor) {
+        if (resourceData?.resource?.resourceUnlocked || userStatus.isInstructor) {
             return 'unlocked';
         }
         if (userStatus.pendingVerification) {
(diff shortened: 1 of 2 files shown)


src/app/pages/details/common/resource-box/resource-box.js
```diff
@@ -45,7 +45,7 @@ function resourceBoxPermissions({
 // is available to the user (instructor version)
 export function instructorResourceBoxPermissions(resourceData, userStatus, search) {
     const resourceStatus = () => {
-        if (resourceData.resourceUnlocked || userStatus.isInstructor) {
+        if (resourceData?.resource?.resourceUnlocked || userStatus.isInstructor) {
```
jivey, line 48: Just trying to expand my knowledge here... was `resourceUnlocked` transformed somewhere? I see `unlocked_resource` in the JSON data you linked in the card... maybe that key gets transformed or I'm looking at the wrong data source.

RoyEJohnson: In general, when I pull data from the CMS I transform it from snake_case to camelCase. Most data is page data, and it's generally loaded using [LoaderPage](https://github.com/openstax/os-webview/blob/d588fa21f499e0ffe7909d03f0fd661ebb159ce3/src/app/components/jsx-helpers/loader-page.js#L12), which pulls the data and processes it.

There are other fetches, and most of them call `camelCaseKeys`.

jivey (approved):

PR #2404, opened 2023-05-23
fix analytics events

TomWoodward:
- aligned donation nudge name
- added logic that waits to configure GA tracker until content is loaded, and set user type tags on the config
- add logic that sets content tags when bookish pages are loaded (will be adding this to other page types like blog as well)
- remove `content-id` parameter from analytics, it turns out it was not really helpful and not worth the confusion of having it in there
- identify featured resources list as being instructor specific
- remove categories parameters from select-content events. the way they were being tracked was not really working, replaced with "tags" which are formatted in a way to make it easy to use a "contains" filter on them, but will not be able to be used as group_by in any meaningful way in GA, maybe another tool will be able to figure it out
- fixed tracking on partner links, i think maybe google doesn't track outbound links if they open target=_blank

### DIFF
--- a/src/app/helpers/use-document-head.js
+++ b/src/app/helpers/use-document-head.js
@@ -59,7 +59,7 @@ function getPageDescriptionElement() {
 }
 
 export function getPageDescription() {
-    return getPageDescriptionElement().getAttribute('content');
+    return getPageDescriptionElement()?.getAttribute('content');
 }
 
 export function setPageDescription(description) {
--- a/src/app/helpers/use-document-head.js
+++ b/src/app/helpers/use-document-head.js
@@ -54,9 +54,17 @@ export function useNoIndex(controlsHeader) {
     );
 }
 
+function getPageDescriptionElement() {
+    return document.querySelector('head meta[name="description"]');
+}
+
+export function getPageDescription() {
+    return getPageDescriptionElement().getAttribute('content');
+}
+
 export function setPageDescription(description) {
-    const descriptionEl = document.querySelector('head meta[name="description"]');
     const defaultDescription = 'Access our free college textbooks and low-cost learning materials.';
+    const descriptionEl = getPageDescriptionElement();
 
     if (descriptionEl) {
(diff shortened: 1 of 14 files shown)


src/app/pages/partners/partner-details/partner-details.js
```diff
@@ -171,14 +171,14 @@ function PartnerDetails({model}) {
...
         <div
-            className="partner-details" onClick={(e) => e.stopPropagation()}
+            className="partner-details"
             ref={ref}
         >
             <div className="sticky-region">
                 <Synopsis {...{model, icon, partnerLinkProps}} />
                 <TabGroup {...{labels, selectedLabel, setSelectedLabel}} />
             </div>
-            <div className="scrolling-region boxed" onClick={(e) => e.stopPropagation()}>
```
TomWoodward, line 181: @RoyEJohnson i have no idea what these `stopPropagation` were here for, but they make it real hard to integrate with what is happening. do you know what their purpose is?

RoyEJohnson: I don't know specifically, but I would do this if clicks were being captured on something that contained these things. It's been in there since Nov 2020. My guess is it was event-tracking related and it will not introduce problems to take it out.

src/app/helpers/use-document-head.js
```diff
@@ -70,11 +79,21 @@ function setPageTitleAndDescription(title='OpenStax', description) {
     document.title = title.includes('OpenStax') ? title : `${title} - OpenStax`;
 }
 
+// eslint-disable-next-line complexity
 export function setPageTitleAndDescriptionFromBookData(data={}) {
     const meta = camelCaseKeys(data.meta || {});
     const defaultDescription = data.description ?
         htmlToText(data.description) : '';
 
+  console.log(data);
```
RoyEJohnson: Need to remove this?

RoyEJohnson (approved):

PR #2409, opened 2023-06-02
Load menus from the CMS

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/1716


Dantemss (approved):
I haven't looked at how the menu data is structured in the CMS, but I delegate that to QA. The code makes sense to me.

PR #2415, opened 2023-06-06
Add overlay boxes to homepage

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2261
For: https://github.com/openstax/openstax.org/issues/2254


jivey (commented):
Should we adjust the underlying background at narrower sizes? The gray background looks a little strange especially.

image

jivey (approved):
Code change looks good, I'll let you decide if the mobile background thing is important or should be a new card.

PR #2433, opened 2023-07-11
Add redirect location to renewal form

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2275

### DIFF
--- a/src/app/components/shell/microsurvey-popup/adoption-content.js
+++ b/src/app/components/shell/microsurvey-popup/adoption-content.js
@@ -60,6 +60,9 @@ function useDismissalCookie() {
 function AdoptionContentBase({children, disable}) {
     const {userModel} = useUserContext();
     const {first_name: name} = userModel || {};
+    const {pathname} = useLocation();
+    const href = `${window.location.origin}${pathname}`;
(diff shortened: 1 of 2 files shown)


src/app/components/shell/microsurvey-popup/adoption-content.js
```diff
@@ -60,6 +60,9 @@ function useDismissalCookie() {
 function AdoptionContentBase({children, disable}) {
     const {userModel} = useUserContext();
     const {first_name: name} = userModel || {};
+    const {pathname} = useLocation();
+    const href = `${window.location.origin}${pathname}`;
+    const renewalFormHref = `/renewal-form?from=popup&href=${encodeURIComponent(href)}`;
```
TomWoodward, line 65: the meaning of the `href` param isn't really clear to me, would `referrer` be more correct? or `from_url`?

TomWoodward (changes requested):

src/app/components/shell/microsurvey-popup/adoption-content.js
RoyEJohnson: I have changed it to `returnTo`, as it is the url to return to after submitting the form.

TomWoodward (approved):

PR #2434, opened 2023-07-12
rip out ga3 analytics

TomWoodward:


src/app/helpers/tag-manager.js
```diff
@@ -27,59 +28,14 @@ window.dataLayer = window.dataLayer || [];
     }
 })(window, document, 'script', 'dataLayer', tagManagerID);
 
-/*
```
TomWoodward, line 30: i moved this deleted logic into the analytics helper so it could be shared with rex.

RoyEJohnson (approved):

PR #2459, opened 2023-08-17
Zero search results

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2298


src/app/pages/blog/pinned-article/pinned-article.js
```diff
@@ -4,7 +4,7 @@ import useBlogContext from '../blog-context';
 import Section from '~/components/explore-page/section/section';
 import './pinned-article.scss';
 
-export default function PinnedArticle({subhead}) {
+export default function PinnedArticle({subhead=null}) {
```
jivey, line 7: Suggestion: I might be missing if null is needed, but I think you can leave this undefined here

jivey (approved):

PR #2460, opened 2023-08-22
Promote Assignable on book details & Subjects pages

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2200

### DIFF
--- a/src/app/components/book-tile/book-tile.js
+++ b/src/app/components/book-tile/book-tile.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import AssignableBadge from '~/components/assignable-badge/assignable-badge';
 import { useToggle } from '~/helpers/data';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCaretUp } from '@fortawesome/free-solid-svg-icons/faCaretUp';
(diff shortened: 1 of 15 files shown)


src/app/components/book-tile/book-tile.js
```diff
@@ -117,11 +118,14 @@ function useBookState(id) {
...
-    const state = useBookState(book.id);
-    const classes = cn({'book-tile': true, 'coming-soon': state === 'coming_soon'});
+    const info = useBookInfo(book.id);
+    const comingSoon = info?.state === 'coming_soon';
+    const assignable = info?.assignable_book === true;
+    const classes = cn({'book-tile': true, 'coming-soon': comingSoon, assignable});
 
     return (
         <div className={classes}>
+            {assignable && <AssignableBadge />}
```
jivey, line 128: Should we make the badge clickable too? When I was testing it, it seemed like the entire cover area should be clickable (though may have to wrap in a separate `a` tag so the tooltip hover still works.)

jivey (changes requested):
Looking good, left a note about making the badge clickable.

src/app/components/book-tile/book-tile.js
RoyEJohnson: It works.

jivey (approved):

PR #2462, opened 2023-08-28
Hotfix 2.110.2

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2313

### DIFF
--- a/src/app/components/jsx-helpers/raw-html.tsx
+++ b/src/app/components/jsx-helpers/raw-html.tsx
@@ -6,7 +6,7 @@ function activateScripts(el: HTMLElement) {
     const processOne = (() => {
         const s = scripts.shift();
 
-        if (!s?.textContent) {
+        if (!s) {
             return;
         }
         const newScript = document.createElement('script');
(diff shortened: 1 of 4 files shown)


src/app/components/jsx-helpers/raw-html.tsx
```diff
@@ -16,7 +16,9 @@ function activateScripts(el: HTMLElement) {
 
         Array.from(s.attributes)
             .forEach((a) => newScript.setAttribute(a.name, a.value));
-        newScript.appendChild(document.createTextNode(s.textContent));
+        if (s.textContent) {
```
TomWoodward, line 19: you should never hit this condition because of the other one you added above right?

TomWoodward (approved):
looks fine, i think one of those conditions might be redundant though

src/app/components/jsx-helpers/raw-html.tsx
RoyEJohnson: That is what I had to split up. I was testing for both `s` and `textContent` at the top, but the `textContent` test had to be moved down.

PR #2467, opened 2023-09-11
Fix language links on specific subjects pages

RoyEJohnson:
For: https://github.com/openstax/os-webview/issues/2466
and: https://github.com/openstax/openstax.org/issues/2327


TomWoodward (approved):
I don't 100% know what I'm looking at but it looks ok to me

PR #2468, opened 2023-09-11
Fix flashing red boxes when switching languages

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2324


src/app/pages/subjects/new/import-subjects-listing.js
```diff
@@ -0,0 +1,3 @@
+import SubjectsListing from './subjects-listing';
+
+export default SubjectsListing;
```
jivey, line 3: I'm not following why this is needed vs importing directly from subjects-listing, is there a circular dependency issue?

src/app/pages/subjects/new/subjects-listing.tsx
```diff
@@ -0,0 +1,78 @@
...
+import useSubjectsContext, {SubjectData} from './context';
+import useSubjectCategoryContext from '~/contexts/subject-category';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faArrowRight} from '@fortawesome/free-solid-svg-icons/faArrowRight';
+import {useLocation} from 'react-router-dom';
+import './subjects-listing.scss';
+
+export default function SubjectsListing() {
+    const {subjects} = useSubjectsContext();
+    const names = Reflect.ownKeys(subjects) as Array<string>;
```
jivey, line 11: This is interesting, I actually haven't seen Reflect before, I'm guessing this is better than Object.keys?

src/app/pages/subjects/new/import-subjects-listing.js
RoyEJohnson: I cannot dynamically import TS files, so I make a little JS importer for each TS file I need to import using `import()`. There are probably webpack plugins and whatnot that can make it import TS, but this was pretty easy.

src/app/pages/subjects/new/subjects-listing.tsx
RoyEJohnson: Arguably better [in some ways](https://blog.greenroots.info/javascript-why-reflect-apis). Reflect is the ES6 package that collects all the introspection stuff as static functions rather than Class methods (there's no instantiable Reflect object).

jivey: Thanks for the explanation!

jivey (approved):

PR #2477, opened 2023-09-19
Add upcoming webinars section

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2337


src/app/pages/webinars/webinar-cards/webinar-grid-section.tsx
```diff
@@ -0,0 +1,37 @@
...
+    children
+}: React.PropsWithChildren<Args>) {
+    const filteredWebinars = React.useMemo(
+        () => webinars.filter(filter).slice(0, limit),
+        [webinars, filter, limit]
+    );
+
+    if (webinars.length === 0) {
+        return null;
+    }
```
bethshook, line 28: could you instead do this check in the parent components and just not render this if webinars is empty?

bethshook (approved):

src/app/pages/webinars/webinar-cards/webinar-grid-section.tsx
RoyEJohnson: I could, but then it's a check in multiple places. I think it's better to have functions handle degenerate cases so the callers don't have to worry about it.

PR #2484, opened 2023-10-05
Show give dialog when downloading PDF from new Subjects

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2352

Added code is pulled almost straight from [the related code on the Details page](https://github.com/openstax/os-webview/blob/main/src/app/pages/details/common/get-this-title-files/options.js#L130)


jivey (approved):
Looks good, one suggestion: because `isMobileDisplay` is only seeing the width at render time, you could have a user start wide and then narrow, so I think it would be good to fix the button wrapping so that it looks right regardless of width:

mobile

PR #2490, opened 2023-10-20
Show give dialog when going to view online

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2361


jivey (approved):
A lot of this looks like the typescript conversion without functionality changes, so I skimmed a bit, but let me know if you have any specific areas that could use a closer look. 

I'm also wondering if the modal should be shown on mobile. I didn't see anything on the card mentioning it, but just wanted to highlight it in case that is important.

RoyEJohnson:
The second and fourth commits are where the new functionality was added. The first and third are largely refactoring code so I can use them for the new functionality, plus migrating to TypeScript, so if you felt like you should check the new code more closely, those would be the commits to look at (and they're smaller).

PR #2498, opened 2023-10-31
Add Polish to language selectors

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2349


src/app/components/language-selector/language-selector.js
```diff
@@ -8,9 +8,12 @@ import './language-selector.scss';
 // You can't use a variable for id
```
bethshook, line 8: are you sure? I think we pass in a prop as the message ID in a few places on rex

bethshook (approved):

src/app/components/language-selector/language-selector.js
RoyEJohnson: It works when I run it, but Jest fails
`    SyntaxError: /Users/rej2/Documents/GitHub/os-webview/src/app/components/language-selector/language-selector.js: [React Intl] Messages must be statically evaluate-able for extraction.
`

RoyEJohnson: I've reworked it to use Intl.DisplayNames for language names.

PR #2505, opened 2023-11-30
Only use donation experiment for PDF download

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2361
View online download dialog always uses the control variant


TomWoodward (changes requested):
can we extract the experiment code to the [give-before-pdf.tsx](https://github.com/openstax/os-webview/pull/2505/files#diff-9d66a576754ac7abc79a138f71b3cfb53dfa1f803806cf98704b3318be80500e) element, maybe just support passing props `header_subtitle` and `give_link` to the CommonElements component as overrides

i would just prefer not to have implementation awareness and branch-boundy logic in a shared component if we can put the same code one level up when we know what we're doing

TomWoodward (approved):

PR #2506, opened 2023-12-01
Add osano

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2364

I'm not trying to make you review everything I do, Tom, it's just that you're the one who knows these things.


TomWoodward (changes requested):
- indentation is a little messed up in the index.html
- may want to add the comment that explains the styles in the index.html (they give the app time to hide the cookie icon so it doesn't flash during page load)
- why are the extra .osano-cm-window styles necessary? i had some css-reset issues with them as well, i'm just wondering what your problem was in this case. the filename is also `osano` not `osana`
- we need to add the ui-components library and footer link

RoyEJohnson:
The ui-components library seems like a heavy add for osweb, which doesn't use StyledComponents (and in fact doesn't use React, but Preact). I made a similar toggler for the footer.

src/app/components/shell/footer/footer.js
```diff
@@ -79,6 +78,27 @@ function Footer({
...
+    );
+
+    if (osano === undefined || osano.cm.mode === 'debug') {
+        document.body.classList.remove('hide-osano');
+        return null;
+    }
+
+    document.body.classList.add('hide-osano');
+    return (
+        <a href='//' role='button' onClick={showOsanoDrawer}>Manage cookies</a>
```
TomWoodward, line 98: can you use an actual button here?

TomWoodward (changes requested):
can you explain why the same popup shows up at the top of the screen in osweb when it shows up at the bottom of the screen in rex? if we need the extra styles then we need them, but i'd like to understand why

src/app/components/shell/footer/footer.js
RoyEJohnson: It would require undoing all the styling for button.

RoyEJohnson: It looks like Osano brings in its own default styling that includes `top: 0`
image
Rex sets bottom, which seems to overrule the top.
```
.osano-cm-dialog--type_bar.osano-cm-dialog--position_bottom {
    bottom: 0;
}
```

RoyEJohnson: Position weirdness:
In osweb, the osano-cm-window is position: fixed; bottom: 0; top: auto;
Within it, the dialog is position: relative.

In rex, osano-cm-window is position: absolute; top: 0; height: 0;
Within it, the dialog is position: fixed; bottom: 0.

I can take out the styling for the osano-cm-window and change the dialog to fixed/bottom.

RoyEJohnson: Using a button (styled as a button):
image

TomWoodward: re:
> This is due to a sitewide rule setting dialog position.

is this something that we should make more selective? if we changed it would we need all the extra styling?

RoyEJohnson: I found that I have an already-defined `.footer-dialog` style; I have piggybacked `.osano-cm-dialog` on it and added a few other lines for a smaller overall osano styling footprint.

TomWoodward (approved):

PR #2515, opened 2024-01-02
Make put away icon keyboard operable

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2386


src/app/components/put-away/put-away.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <span
+            className='put-away'
+            aria-label={ariaLabel}
+            role='button'
+            onClick={onClick}
+            tabIndex={0}
+            onKeyDown={closeUsingKeyboard}
+            data-nudge-action='dismissed'
+        >
+            &times;
```
jivey, line 27: Don't you want to keep the FontAwesome icon for this?

src/app/components/shell/lower-sticky-note/lsn-content.js
```diff
@@ -23,7 +23,7 @@ function NoteWithImage({bannerInfo}) {
 
 function NoteWithoutImage({bannerInfo}) {
     return (
-        <div className="content">
+        <div className="content" role='marquee' aria-label='sticky note'>
```
jivey, line 26: Would a `status` role be more appropriate here, because the banner text doesn't (seem to) change? I'm not sure what is normal to use here, but the marquee docs suggest it should be for live, updating content.

jivey (changes requested):
Looks good just wondering about the role and font awesome icon.

src/app/components/put-away/put-away.tsx
RoyEJohnson: I had some with FontAwesome and some with `&times`. They should be consistent, so I picked `&times`.

src/app/components/shell/lower-sticky-note/lsn-content.js
RoyEJohnson: I had a hard time picking. Both status and marquee are live regions. It seemed more of an announcement than a status message to me.

RoyEJohnson: Dialog was the other contender.

jivey: Yeah that's a good point about live region, which it doesn't seem to be. `dialog` now seems like the most appropriate candidate to me.

src/app/components/put-away/put-away.tsx
jivey: I'd personally pick the FontAwesome to standardize on, because it's less prone to font rendering differences/issues. I've found the svgs to be much easier to keep centered in buttons and other ui elements.

jivey (approved):

PR #2520, opened 2024-01-17
Make menus keyboard operable

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2385

### DIFF
--- a/src/app/components/shell/header/menus/main-menu/dropdown/dropdown.js
+++ b/src/app/components/shell/header/menus/main-menu/dropdown/dropdown.js
@@ -4,16 +4,25 @@ import useDropdownContext from '../../dropdown-context';
 import useWindowContext from '~/contexts/window';
 import {isMobileDisplay} from '~/helpers/device';
 import {treatSpaceOrEnterAsClick} from '~/helpers/events';
+import {useLocation} from 'react-router-dom';
 import cn from 'classnames';
 import './dropdown.scss';
 
+// Using ARIA Disclosure pattern rather than Menubar pattern
+// because menubar requires complexity that is not necessary
+// for ordinary website navigations, per
+// https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/
+
 export function MenuItem({label, url, local}) {
     const {innerWidth: _} = useWindowContext();
+    const urlPath = url.replace('/view-all', '');
+    const {pathname} = useLocation();
 
     return (
         <RawHTML
             Tag="a" html={label}
-            href={url} tabIndex={0} data-local={local} role="menuitem"
+            href={url} tabIndex={0} data-local={local}
+            {...(urlPath === pathname ? {'aria-current': 'page'} : {})}
         />
     );
 }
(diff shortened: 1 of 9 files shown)


TomWoodward (approved):

Dantemss:
@TomWoodward the menu doesn't really follow the spec. Probably up to you or Erika to decide how closely we need to follow it?

Both main and submenu:
- Menu auto-opens on hover/focus instead of waiting for space/enter/down arrow/up arrow.
- No wraparound for arrow navigation when you reach the first/last item. Up arrow also goes back to the main nav from the submenu.
- Home and end scroll the page instead of the menu. Now I'm wondering if I need to add preventDefault() for these on my PR.
- Pressing the first letter of a label for keyboard navigation doesn't work
- Esc closes the submenu but does not focus the main menu item (that would auto-open the menu again anyway since it has that behavior)

Submenu only:
- Space scrolls the page a bit, does not select the menu item

Dantemss (changes requested):
I'd say we should fix the up/down arrows on mobile and probably make space also select the submenu item. The rest could probably be debated.

RoyEJohnson:
Note the [comment](https://github.com/openstax/os-webview/pull/2520/files#diff-56a7e138b86498367b98baebe98fce08c7ffc3408866a200496ada19b98f6634R11) where I point out that I removed the menubar roles and went with the (simpler) Disclosure pattern instead. The link there says most sites don't need the full features of the menubar.
Intended navigation is tab, but I added arrows to allow moving between main menu items. I should make the arrow behavior sensible in all contexts.

Dantemss:
OK I think most of what I said still applies, except the arrows are supposed to do different things from what I said in the disclosure pattern. I'll re-test the tabbing though since that wasn't in the other one.

Dantemss:
Tabbing seems to work as expected in both Desktop and mobile. And the disclosure pattern doesn't need to do anything when you type a letter (no search by the first letter of the labels).

Dantemss:
The only thing missing I think is that the space bar is not activating links when it should, and instead is scrolling the page (it does enable the dropdowns in mobile mode).

Dantemss (changes requested):
The only thing missing I think is that the space bar is not activating links when it should, and instead is scrolling the page (it does enable the dropdowns in mobile mode).

Dantemss:
Also the desktop version is not quite disclosure pattern because of the auto-open on hover/focus but that is acceptable I think?

Dantemss (commented):
Mobile menu still needs click on space

Dantemss:
Maybe that's not a thing? How do screen readers work on mobile anyway?

Dantemss:
I suppose it would be helpful if you somehow had a tiny screen reader screen.

RoyEJohnson:
I think the assumption is that you _can_ hook up an external keyboard to a mobile device.

TomWoodward:
or you might just have a tiny screen like a netbook or have it really zoomed in or something

Dantemss (approved):

PR #2524, opened 2024-01-24
Don't apply sitewide dialog styles to Osano dialogs

RoyEJohnson:
I don't know if we had a way to test Osano outside of production?
There are two commits:
The first excludes Osano dialogs from default dialog styling. It keeps styling that was specific to osano-cm-dialog, but we can take that out, too, if needed.
The second drops all dialog styling outside of #app (which excludes all the osano-specific styling as well, so I deleted those styles).

### DIFF
--- a/src/app/components/dialog/dialog.scss
+++ b/src/app/components/dialog/dialog.scss
@@ -2,21 +2,22 @@
 
 dialog,
 [role="dialog"] {
-    background-color: ui-color(white);
-    border: 0;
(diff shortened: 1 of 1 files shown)


TomWoodward (approved):

TomWoodward (changes requested):
Actually I didn't realize that last part was still in there, can we remove that too?

TomWoodward (approved):

PR #2526, opened 2024-01-26
Fix Kinetic page canonical link

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2397

### DIFF
--- a/src/app/components/shell/router.js
+++ b/src/app/components/shell/router.js
@@ -36,7 +36,8 @@ function ImportedPage({name}) {
             function Loading({error, pastDelay, retry}) {
                 if (error) {
                     if (error.code === 'MODULE_NOT_FOUND') {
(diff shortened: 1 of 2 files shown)


src/app/components/shell/router.js
```diff
@@ -36,7 +36,8 @@ function ImportedPage({name}) {
             function Loading({error, pastDelay, retry}) {
                 if (error) {
                     if (error.code === 'MODULE_NOT_FOUND') {
-                        return <FallbackToGeneralPage name={name} />;
+                        return pathname.endsWith('/') ? <FallbackToGeneralPage name={name} />
```
jivey, line 39: Can we add a comment about the case this is handling?

jivey (approved):
This change looks okay to me, but a comment about the fallback case might help me not get confused if I look at this in the future 😄

PR #2532, opened 2024-02-06
Use tablist for dynamic features list on homepage

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2400
For: https://github.com/openstax/openstax.org/issues/2401

### DIFF
--- a/src/app/components/tablist/tablist.tsx
+++ b/src/app/components/tablist/tablist.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+// https://react-spectrum.adobe.com/react-aria/useTabList.html
+import {useTabList, useTab, useTabPanel} from 'react-aria';
+import {Item, useTabListState} from 'react-stately';
+import type {Node} from '@react-types/shared/src/collections';
+import './tablist.scss';
+
+type TabsProps = Parameters<typeof useTabListState>[0] &
+    Parameters<typeof useTabList>[0];
+
+function Tabs(props: TabsProps) {
+    const state = useTabListState(props);
+    const ref = React.useRef(null);
+    const {tabListProps} = useTabList(props, state, ref);
+
+    return (
+        <div className={`tabs ${props.orientation || ''}`}>
+            <div {...tabListProps} ref={ref}>
+                {[...state.collection].map((item) => (
+                    <Tab key={item.key} item={item} state={state} />
+                ))}
+            </div>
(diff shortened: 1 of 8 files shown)


src/app/pages/home/features/features.js
```diff
@@ -2,63 +2,49 @@ import React from 'react';
...
-        [data.tab2Features, data.tab2ExploreUrl];
     const bgImage = useOptimizedImage(data.bgImage, 570);
 
     return (
-        <section className="features">
-            <img className="right-bg" src={bgImage} alt />
-            <div className="boxed text-block">
+        <section className='features'>
+            <img className='right-bg' src={bgImage} alt />
+            <div className='boxed text-block'>
```
TomWoodward, line 32: convention is to leave these as double quotes if they're string literals in props https://stackoverflow.com/a/44209432

src/app/components/tablist/tablist.tsx
```diff
@@ -0,0 +1,60 @@
...
+type TabsProps = Parameters<typeof useTabListState>[0] &
+    Parameters<typeof useTabList>[0];
+
+function Tabs(props: TabsProps) {
+    const state = useTabListState(props);
+    const ref = React.useRef(null);
+    const {tabListProps} = useTabList(props, state, ref);
+
+    return (
+        <div className={`tabs ${props.orientation || ''}`}>
```
TomWoodward, line 17: is there a reason you didn't use `react-aria-components` like the example? https://react-spectrum.adobe.com/react-aria/Tabs.html seems like that would do some of the wiring you have above for you?

TomWoodward (changes requested):

src/app/pages/home/features/features.js
RoyEJohnson: One of the few things prettier allows as a configuration, as it turns out.

src/app/components/tablist/tablist.tsx
RoyEJohnson: That was actually code from the tablist page at react-aria. Makes sense that it would be in a package.
It doesn't seem to work the same, so I'll have some figuring out to do.

RoyEJohnson: It looks like react-aria-components don't play nice with Preact.

TomWoodward (approved):

PR #2541, opened 2024-02-21
Set focus color in header for accessible contrast

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2412

### DIFF
--- a/src/app/components/shell/header/header.scss
+++ b/src/app/components/shell/header/header.scss
@@ -1,5 +1,7 @@
 @import 'pattern-library/core/pattern-library/headers';
 
+$focus-color: lighten(os-color(medium-blue), 9%);
+
 .page-header {
     align-items: center;
     background-color: rgba(ui-color(white), 0.98);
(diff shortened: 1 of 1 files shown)


src/app/components/shell/header/header.scss
```diff
@@ -53,6 +55,8 @@
             background-color: os-color(red);
             color: text-color(white);
             padding: 0.5rem $normal-margin;
+            outline: thick solid #1f96a8;
```
jivey, line 58: I'm not sure if os-web has a theme constants file but these colors seem good for reuse?

jivey (approved):

src/app/components/shell/header/header.scss
RoyEJohnson: We have color constants in the pattern library (a separate package).
These colors are pretty close to each other, and also close to the library color `medium-blue`, so I made a lightened version of medium blue that meets the contrast requirements for both and used that. We might want to add it to the pattern library if the focus color issue crops up elsewhere.

PR #2552, opened 2024-03-12
Use accessible carousel on specific-subjects pages

RoyEJohnson:
For: https://github.com/openstax/openstax.org/issues/2413

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
         "boxible": "^1.7.1",
         "classnames": "^2.3.2",
         "esbuild-loader": "^3.0.1",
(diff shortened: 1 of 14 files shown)


jivey (changes requested):
Looking good just noticed some minor css issues.

I'm not sure if this is present on the production version, but the mobile view partially shows the hover-text which makes it look glitchy: 
image

I'm also seeing some positioning issues at certain widths, I wonder if the card css isn't centering the contents?
image
image

jivey (approved):
Thanks for the updates, looking great

PR #2576, opened 2024-04-10
Merge mid-April release to main

RoyEJohnson:
Release candidate for 17 Apr 2024

### DIFF
--- a/src/app/pages/adoption/adoption.js
+++ b/src/app/pages/adoption/adoption.js
@@ -20,6 +20,10 @@ function BookSelectorPage({selectedBooksRef}) {
         [selectedBooks]
     );
     const instructions = 'Select titles you are using even if the edition number is different.';
+    const includeFilter = React.useCallback(
+        (b) => !b.comingSoon,
+        []
+    );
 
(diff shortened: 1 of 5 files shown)


src/app/pages/adoption/adoption.js
```diff
@@ -20,6 +20,10 @@ function BookSelectorPage({selectedBooksRef}) {
         [selectedBooks]
     );
     const instructions = 'Select titles you are using even if the edition number is different.';
+    const includeFilter = React.useCallback(
+        (b) => !b.comingSoon,
+        []
+    );
```
Dantemss, line 26: Is this just for performance reasons? Not sure it would do much vs the inline version. Calling `useCallback` has a cost anyway.

Dantemss (approved):

src/app/pages/adoption/adoption.js
RoyEJohnson: It's for hygiene reasons. I had the dreaded update-creates-update loop. That might have been due only to the default value, but I covered both bases. The performance difference is negligible.

PR #2577, opened 2024-04-10
Fix html validation errors

RoyEJohnson:
[DISCO-194]

[DISCO-194]: https://openstax.atlassian.net/browse/DISCO-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
The primary goal was to differentiate the desktop and mobile dropdown menu IDs. I did that by adding `prefix` to the dropdown context. While I was there, I converted dropdown-context to TS.


src/app/components/shell/header/menus/menus.tsx
```diff
@@ -43,40 +43,44 @@ export default function Menus({open}: {open: string;}) {
...
-                    aria-expanded={active}
-                />
-                <div className='menu-popover-overlay' onClick={clickOverlay}>
-                    <div id='menu-popover' className='menu-popover'>
-                        <div className='menu-title'>Menu</div>
-                        <ul className='no-bullets' onKeyDown={treatSpaceOrEnterAsClick}>
-                            <MainMenuItems />
-                            <UpperMenu />
-                        </ul>
+        <React.Fragment>
```
bethshook, line 46: not sure this is needed here?

bethshook (approved):

src/app/components/shell/header/menus/menus.tsx
RoyEJohnson: There are two DropdownContextProvider units, so I had to wrap them in a Fragment (because of the every-component-must-return-a-single-root-element rule)

PR #2587, opened 2024-04-26
Include Spanish books on adoption form

RoyEJohnson:
[DISCO-190]
I reorganized some code so that I could introduce a different book-fetching option, and ported a few modules to TS while I was there.

[DISCO-190]: https://openstax.atlassian.net/browse/DISCO-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/src/data/all-books.js
```diff
@@ -0,0 +1,566 @@
+export default {
```
bethshook, line 1: can you add formatting to this file?

bethshook (approved):

PR #2590, opened 2024-04-29
PM ticket - Accessbility

mikayBuk:
This pull request has the changes for the research page for the Accessibility ticket. It should change the color of the h1 headers to be black.


src/app/pages/research/research.scss
```diff
@@ -37,6 +37,12 @@
         margin-block: 0 !important;
     }
 
+    // Changes
```
chrisbendel, line 40: I would remove these comments

src/app/pages/research/research.scss
```diff
@@ -37,6 +37,12 @@
...
+    // Adding changes from PM ticker - Mikayla
+    h1{
+        color: #151b2c;
```
chrisbendel, line 43: Instead of making the change here like this (this will affect all <h1> tags), i would make the change here for the specific h1

https://github.com/openstax/os-webview/blob/main/src/app/pages/research/components/header.js#L24

chrisbendel (approved):

chrisbendel:
@RoyEJohnson this should be good to merge and include in whatever next release there is - no rush

RoyEJohnson:
Merge isn't available until check pass and the check seems to be hung up.

chrisbendel:
@RoyEJohnson i see that now - it must be because the PR is coming from a fork, i don't see a way to manually start the checks. is there a way we can just merge this and skip the check for now? its just a one line change and not impactful at all 🤔

RoyEJohnson:
I don't know of a way to bypass the checks, so I made a [new PR](https://github.com/openstax/os-webview/pull/2603) based on this one.

PR #2598, opened 2024-05-07
Submit current URL as tracking_parameters in adoption/interest forms

RoyEJohnson:
[DISCO-264]

[DISCO-264]: https://openstax.atlassian.net/browse/DISCO-264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/tracking-parameters/tracking-parameters.tsx
+++ b/src/app/components/tracking-parameters/tracking-parameters.tsx
@@ -1,0 +1,10 @@
+import React from 'react';
+
+export default function TrackingParameters() {
+    return (
+        <React.Fragment>
+            <input type="hidden" name="application_source" value="OS Web" />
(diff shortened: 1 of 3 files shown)


Dantemss (approved):
Adoption and interest should probably also be .jsx

RoyEJohnson:
With JS files, I don't need to distinguish between JS and JSX. Almost all of them include JSX, but I don't use that suffix.

Dantemss:
I know it works without the extension being .jsx, it's just nitpicking because they use the jsx templating and wouldn't be valid .js files without that. But it makes sense not to change just a couple of files if there are dozens more in the repo.

PR #2600, opened 2024-05-13
Improve test coverage

RoyEJohnson:
[DISCO-273]

[DISCO-273]: https://openstax.atlassian.net/browse/DISCO-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/pages/blog/search-results/use-all-articles.js
```diff
@@ -0,0 +1,28 @@
...
+        fetchFromCMS(slug, true).then((results) => {
+            setAllArticles(
+                uniqBy(results, 'id').map((data) => {
+                    data.heading = data.title;
+                    delete data.subheading;
+                    return blurbModel(camelCaseKeys(data));
+                })
+            );
+        });
+    }, [searchParam]);
```
jivey, line 25: This is from the old code, so maybe all is good, but at a glance I would have thought `search` would be need to be here instead?

test/src/components/breadcrumb.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+// There's not much functionality to inspect, but you should have 100% code coverage
+describe('breadcrumb', () => {
+    it('navigates to new and retraceable links', async () => {
+        render(<Component name="fourth" />);
+
+        const user = userEvent.setup();
+
+        await user.click(screen.getAllByRole('link')[0]);
+        await user.click(screen.getAllByRole('link')[1]);
+        await user.click(screen.getAllByRole('link')[0]);
```
jivey, line 32: Does something different happen when revisiting the same breadcrumb link?

jivey (approved):

src/app/pages/blog/search-results/use-all-articles.js
RoyEJohnson: `searchParam` is what's used in the effect. It is dependent on `search` and not used outside it, so I could create it in the effect and make that the dependency.

PR #2602, opened 2024-05-15
Focus on main when content updates

RoyEJohnson:
[DISCO-283]


bethshook:
what's causing the initial focus on the ad?

bethshook (approved):

PR #2604, opened 2024-05-17
Improve test coverage 2

RoyEJohnson:
[DISCO-280]

[DISCO-280]: https://openstax.atlassian.net/browse/DISCO-280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/src/components/__snapshots__/book-tile.test.tsx.snap
```diff
@@ -31,6 +31,7 @@ exports[`book-tile renders 1`] = `
       >
         <button
           aria-controls="sample-ddm"
+          aria-expanded="false"
```
jivey, line 34: I can't quite tell what change added this, maybe a test changed the open state? Just confirming if it is expected.

jivey (approved):

test/src/components/__snapshots__/book-tile.test.tsx.snap
RoyEJohnson: Previously, the default value of `aria-expanded` was null. Converting to TS made me use boolean, so it's explicit rather than absent.

PR #2618, opened 2024-06-03
Support any browser that supports grid layout

RoyEJohnson:
[DISCO-263]
I simplified the test for supported browsers, reducing the risk of erroneously popping up the warning that the site might not render properly, but increasing the risk that some obscure browser will render badly and there will be no warning.
I don't know of any other likely-helpful features to test for in this regard.

It looks like [eslint-plugin-compat](https://stackoverflow.com/a/54942229) can check for these issues, but it wants node v18.

[DISCO-263]: https://openstax.atlassian.net/browse/DISCO-263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/helpers/device.js
+++ b/src/app/helpers/device.js
@@ -2,46 +2,7 @@ export function isMobileDisplay() {
     return window.innerWidth <= 960;
 }
 
-function browserId() {
-    const ua = window.navigator.userAgent;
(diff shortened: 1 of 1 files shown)


TomWoodward (approved):
seems fine with me if you think the check is sufficient

PR #2628, opened 2024-06-12
Fix carousel next button

RoyEJohnson:
[DISCO-328]

[DISCO-328]: https://openstax.atlassian.net/browse/DISCO-328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (changes requested):
Should we narrow the gap between cards on mobile? It's difficult to see that there's more swipeable content to see
image

I'm also wondering if we should leave the left/right controls visible at smaller screen sizes that don't have touch capability? If you narrow the viewport enough on desktop, you can't navigate because it's looking for a touch event.

RoyEJohnson:
I centered the image and left the left/right controls on at all screen sizes. I don't think the carousel wants to display partial frames to indicate there's more content.

jivey (changes requested):
I think the CSS changes might have messed up the desktop view:

image

RoyEJohnson:
Oops. Thanks for catching that. I think this update gets it right.

PR #2633, opened 2024-06-24
Increase code coverage 4

RoyEJohnson:
[DISCO-355]

[DISCO-355]: https://openstax.atlassian.net/browse/DISCO-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/helpers/use-document-head.ts
+++ b/src/app/helpers/use-document-head.ts
@@ -19,19 +19,6 @@ function setCanonicalPath(newPath: string) {
     return el;
 }
 
-function noindexMeta() {
-    const el = document.createElement('meta');
-    const titleEl = document.querySelector('head title');
-
-    if (titleEl) {
-        el.setAttribute('name', 'robots');
-        el.setAttribute('content', 'noindex');
-        titleEl.parentNode?.insertBefore(el, titleEl.nextSibling);
-    }
-
-    return el;
-}
-
 export function useCanonicalLink(controlsHeader = true, path?: string) {
     const defaultPath = useLocation().pathname;
     const newPath = (path || defaultPath).replace(/\/$/, '');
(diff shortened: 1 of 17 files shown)


src/app/helpers/use-document-head.ts
```diff
@@ -117,6 +93,9 @@ export function setPageTitleAndDescriptionFromBookData(data: BookData = {}) {
         ? htmlToText(data.description)
         : '';
 
+    if (data.bookSubjects && data.bookCategories) {
+        console.info('*** BOOK SUBJECTS AND CATEGORIES FOUND!!!! ');
```
jivey, line 97: Not sure if debug...

test/src/helpers/use-document-head.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+            </MemoryRouter>
+        );
+        expect(
+            document.head.querySelector('link')?.getAttribute('href')
+        ).toContain(canonicalPath);
+    });
+    it('handles fails gracefully if there is no title', () => {
+        const titleEl = document.head.querySelector('title');
+
+        document.head.removeChild(titleEl as Node);
```
jivey, line 241: I see the comment above... maybe a guard here though so it doesn't error if the test or describe block is focused?

test/src/components/shell/use-link-handler.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+        jest.spyOn(linkHelper, 'validUrlClick').mockReturnValue(false);
+
+        render(
+            <MemoryRouter initialEntries={['/']}>
+                <Component />
+            </MemoryRouter>
+        );
+        await user.click(screen.getByRole('link'));
+        expect(notPrevented).toBeCalledWith(true);
+        jest.resetAllMocks();
```
jivey, line 51: `afterEach` on the parent describe block could work for all the reset calls, I think

jivey (approved):
Left some fairly minor suggestions otherwise looks good

src/app/helpers/use-document-head.ts
RoyEJohnson: Whoops.

PR #2638, opened 2024-07-01
Content warning modal

RoyEJohnson:
[DISCO-343]

[DISCO-343]: https://openstax.atlassian.net/browse/DISCO-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/details/common/get-this-title-files/give-before-pdf/content-warning.tsx
+++ b/src/app/pages/details/common/get-this-title-files/give-before-pdf/content-warning.tsx
@@ -1,12 +1,13 @@
 import React from 'react';
 
 export default function ContentWarning({
-    link, track, close, onDownload
+    link, track, close, onDownload, warning
 }: {
     link: string;
     track: string | undefined;
     close: () => void;
     onDownload: (event: React.MouseEvent) => void;
+    warning: string;
 }) {
     const closeAfterDelay = React.useCallback(
         (event: React.MouseEvent) => {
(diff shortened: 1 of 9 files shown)


src/app/pages/details/common/get-this-title-files/give-before-pdf/content-warning.tsx
```diff
@@ -0,0 +1,37 @@
...
+            <div className="text-content">
+                {warning}
+            </div>
+            <a
+                href={link}
+                {...(track ? {'data-track': track} : {})}
+                onClick={closeAfterDelay}
+                className="btn go-to"
+            >
+                Go to your file
```
jivey, line 33: Could we make this work like the give modal does, so it would say "Go to your book" if that was the resource clicked? If not, maybe "Continue"?

jivey (changes requested):
Looking good, small concern about resource wording.

src/app/pages/details/common/get-this-title-files/give-before-pdf/content-warning.tsx
RoyEJohnson: I made it use "book" for "View online" but left it "file" for PDF. Should it be "book" for PDF, too?

jivey: To me it makes sense to match the give modal so I think file is fine?

jivey (approved):

PR #2645, opened 2024-07-24
remove old homepage

TomWoodward:
[DISCO-435]
the way the router works, if the old homepage (which was defined in the ui as `home`) is removed, it'll then search for a page with the slug `home`, and render it normally.

so to publish the new homepage all we have to do is make sure its slug is `home` and delete the previous code

[DISCO-435]: https://openstax.atlassian.net/browse/DISCO-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TomWoodward:
do not merge until qa is done for previous deployment (we need to deploy the flex page editor and create the new page before we deploy this to use it as the homepage)

mwvolo:
@TomWoodward can this be deployed to dev for testing social metadata stuff (I'd feel more confident removing the slug for `openstax-homepage` in the CMS code if so)? Or do we want to wait on that?

TomWoodward:
i think brittany wanted to qa on dev what will be in the first release, so we probably want to wait on that

mwvolo (approved):

PR #2646, opened 2024-07-24
move layouts out of shell

TomWoodward:
this doesn't functionally change anything. it moves the default layout out of the main shell and into a new `/layouts` directory and allows for the possibility of a `FlexPage` specifying an alternate layout. i moved a few things out of the layout and into the router that we absolutely want all the time (skip to content and title announcement)

### DIFF
--- a/src/app/layouts/default/footer/footer.js
+++ b/src/app/layouts/default/footer/footer.js
@@ -23,7 +23,7 @@ function Footer({
         facebookLink, twitterLink, linkedinLink
     }
 }) {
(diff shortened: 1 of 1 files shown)


RoyEJohnson (commented):
This makes sense. Just the test of the shell component to fix up.

TomWoodward:
@RoyEJohnson finally got the tests passing. there is a lot of warning output that makes working on the tests very confusing, but i guess it was all pre-existing?

RoyEJohnson (approved):
Yes, there are a lot of data requests that the fetch-mocker doesn't handle, but the associated tests don't care. That's all pre-existing. They could be addressed by mocking fetch for those tests.

PR #2648, opened 2024-07-25
Increase code coverage 5

RoyEJohnson:
[DISCO-431]

[DISCO-431]: https://openstax.atlassian.net/browse/DISCO-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/models/book-toc.ts
```diff
@@ -3,13 +3,10 @@ import {cnxFetch} from './table-of-contents-html';
 import memoize from 'lodash/memoize';
 
 export function bookToc(slug: string) {
-    if (!slug) {
```
jivey, line 6: Is this guaranteed to not be empty, or maybe prefer to have the throw in `cmsFetch`?

src/app/models/book-toc.ts
```diff
@@ -3,13 +3,10 @@ import {cnxFetch} from './table-of-contents-html';
...
-        return Promise.reject(new Error('No slug to fetch'));
-    }
     return cmsFetch(slug)
         .then((bi) => {
-            const isRex = Boolean(bi.webview_rex_link);
-            const webviewLink = isRex ? bi.webview_rex_link : bi.webview_link;
+            const isRex = true;
```
jivey, line 8: I see that this controls which paths to use but I'm not following why it can always be true

src/app/models/book-toc.ts
RoyEJohnson: The [only place it is called](https://github.com/openstax/os-webview/blob/main/src/app/pages/errata-form/form/ErrorLocationSelector/toc-selector.js#L93) tests for `slug`.

src/app/models/book-toc.ts
RoyEJohnson: This is leftover from the transition to Rex. The non-Rex option was CNX, which no longer exists.
There's no longer a `webview_link` field in the CMS. I need to go through the codebase and remove `isRex` generally.

jivey (approved):

PR #2656, opened 2024-07-31
fuss with font size on flex pages

TomWoodward:
increases font size in buttons, removes heading size overrides so they use the default heading sizes, only use different heading size within cards where the h2 is really just too big

### DIFF
--- a/src/app/pages/flex-page/blocks/RichTextBlock.scss
+++ b/src/app/pages/flex-page/blocks/RichTextBlock.scss
@@ -11,17 +11,8 @@
     margin-top: 0;
   }
 
-  h1 {
-    @include title-font(4.8rem);
-  }
-  h2 {
-    @include title-font(2.6rem);
-  }
-  h3 {
-    @include title-font(1.8rem);
-  }
-  h4, h5, h6 {
-    @include title-font(1.6rem);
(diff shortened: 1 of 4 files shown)


TomWoodward:
this is based on feedback from CS on the button font mostly

src/app/pages/flex-page/blocks/RichTextBlock.scss
```diff
@@ -11,17 +11,8 @@
...
-  h2 {
-    @include title-font(2.6rem);
-  }
-  h3 {
-    @include title-font(1.8rem);
-  }
-  h4, h5, h6 {
-    @include title-font(1.6rem);
+  h5, h6 {
+    @extend %text-h4;
```
RoyEJohnson, line 15: I had forgotten about the `%text-` stuff! Same as `@include set-font(h4)` if you prefer `@include`.

RoyEJohnson (approved):

src/app/pages/flex-page/blocks/RichTextBlock.scss
TomWoodward: i copied this out of the pattern library sass, i don't really care either way

PR #2657, opened 2024-07-31
tweaks to flexpage styling

TomWoodward:
increases base line height on flex-pages and adds dynamic text color for heroes based on background. used in this spike for the foundation pages
localhost_3000_supporters-new_


TomWoodward:
@RoyEJohnson can you check this one out

src/app/pages/flex-page/blocks/RichTextBlock.scss
```diff
@@ -17,6 +17,7 @@
 
   p {
     @include body-font(1.6rem);
+    line-height: 2.5rem;
```
RoyEJohnson, line 20: This is the same as `@include set-font(body-regular)`, which is the usual text block stuff on osweb.
There's also `@include set-font(body-large)` which has 1.8rem font and line-height of 3rem.

RoyEJohnson (approved):
The color package seems handy.

PR #2658, opened 2024-08-06
Fix menu flicker

RoyEJohnson:
[DISCO-433]
Separated Header and Footer from the changing content in DefaultLayout so that the whole thing isn't destroyed and re-created on every navigation.
Currently, every ImportedPage is a DefaultLayout, so it's pretty easy. Depending on what future layouts are like, we may want to make the Header/Footer conceptually distinct from the layout.

[DISCO-433]: https://openstax.atlassian.net/browse/DISCO-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson:
This is currently installed on Dev.

TomWoodward (changes requested):
this solution nests all pages inside the default layout navs. see https://dev.openstax.org/b2s-fall-24/

RoyEJohnson:
I had missed the way Fallback is done, so thought everything was a DefaultLayout somehow.
Updated code (on Dev) seems to work pretty well. Still flickers between home page and other pages, but otherwise the wrapper generally persists. The b2s page is not wrapped.

RoyEJohnson:
I made it so that ImportedPage always applies the Layout. The content components set parameters for the Layout component, but don't apply layouts themselves. This has made the menus stable except when coming to and from the home page. I don't know what causes it to redraw menus.

TomWoodward (changes requested):
is there a bug here where if you navigate to a page that specifies a layout, and then you navigate back to a page that doesn't specify a layout, it won't switch back to the default?

src/app/components/shell/router.js
```diff
@@ -31,41 +31,66 @@ const Error404 = loadable({
...
+            }
+        }),
+        [name, loading]
+    );
+}
+
+function ImportedPage({name}) {
+    const {pathname} = useLocation();
+    const [layoutName, setLayoutName] = React.useState('default');
+    const [layoutData, setLayoutData] = React.useState(undefined);
```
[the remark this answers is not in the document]
RoyEJohnson: For non-obvious reasons, it doesn't work to make a single `layoutParameters` state object and access its members for `layoutName` and `layoutData`. It causes an update loop when loading a page with non-default layout (e.g., b2s-fall-24)
Navigating from b2s-fall-24 to the home page via the staxly logo link works as expected. It looks like it might not switch to default if it falls back to a general page, so I have added a line to handle that.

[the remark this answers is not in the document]
RoyEJohnson: It looks like `data` needs a little attention to keep it from having unnecessary reloads, so I tried to optimize that.
This branch is on Dev now. There is still some rebuilding of menus when going between some of the pages (K12 and Webinars under "What we do" for example), but the menus persist for most normal page transitions.

RoyEJohnson:
@TomWoodward I have pulled the Layout up another level to wrap around the Routes. I also extracted it to a Context. The higher level and some careful handling of data changes keep the menus from re-rendering.
It is ready to review.

src/app/components/shell/router-helpers/fallback-to.js
```diff
@@ -29,16 +23,14 @@ export default function FallbackTo({name}) {
     // i think page-data-utils:fetchFromCMS would have to be updated
     // to do something special on a 404 status
     if ('error' in data) {
-        return <Layout name="default"><Error404 /></Layout>;
+        return <Error404 />;
```
TomWoodward, line 26: the 404 page should probably always be shown in the default layout, should you call `setLayoutParameters` to change it to `default` here?

src/app/components/shell/router-helpers/fallback-to.js
```diff
@@ -29,16 +23,14 @@ export default function FallbackTo({name}) {
...
     }
 
     if (['pages.FlexPage', 'pages.RootPage'].includes(data.meta.type)) {
-        return <Layout data={data} name={data.layout[0]?.type || 'default'}>
-            <FlexPage data={data} />
-        </Layout>;
+        setLayoutParameters({data, name: data.layout[0]?.type || 'default'});
```
TomWoodward, line 30: i think its usually considered to be a no-no to call a state setter in a render function, i think you're supposed to put it in a `useEffect` handler so it only calls when the inputs change?

src/app/components/shell/router.js
```diff
@@ -31,41 +32,51 @@ const Error404 = loadable({
...
+    const {setLayoutParameters} = useLayoutContext();
+
+    return React.useMemo(
+        () => loadable({
+            loader: () => import(`~/pages/${name}/${name}`),
+            loading,
+            render(loaded, props) {
+                const Component = loaded.default;
+
+                setLayoutParameters({name: 'default'});
```
TomWoodward, line 69: this may cause flicker when navigating between two non-default layout pages, but to be fair i don't think we have that situation in reality right now

src/app/contexts/layout.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import buildContext from '~/components/jsx-helpers/build-context';
+import loadable from 'react-loadable';
+import LoadingPlaceholder from '~/components/loading-placeholder/loading-placeholder';
+
+// Webpack wasn't able to make the dynamic strings work in the Context value.
```
TomWoodward, line 6: sometimes webpack does an unfortunate thing where if you use a variable in an import it loads the entire directory, this is happening in rex with the intl polyfil and its horrible

src/app/contexts/layout.tsx
```diff
@@ -0,0 +1,52 @@
...
+type LayoutName = 'default' | 'landing';
+const loaders = {
+    'default': () => import('~/layouts/default/default'),
+    landing: () => import('~/layouts/landing/landing')
+};
+
+function useContextValue() {
+    const [layoutName, setLayoutName] = React.useState<LayoutName>('default');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const [layoutData, setLayoutData] = React.useState<any>(undefined);
```
TomWoodward, line 16: i commented on this before but i forget, what was the issue with combining these two states into a 
```
React.useState<{ name: LayoutName; data: any;}>({name: 'default', data: undefined}); 
```
kind of thing?

TomWoodward (changes requested):

src/app/components/shell/router-helpers/fallback-to.js
RoyEJohnson: Ok. I also set defaults for `setLayoutParameters` so that without arguments it sets default layout.

src/app/components/shell/router.js
RoyEJohnson: I think if it gets to the `render` there, it's a default layout page, right?

src/app/contexts/layout.tsx
RoyEJohnson: Need to be able to detect when a change of value isn't a real change, to prevent flicker or render cascades.
I was able to use `useReducer` instead of `useState` and have it work.

TomWoodward: ah right, because updating it to a new object using the same layout/params would be a different strict equality check, but with just a string and `undefined` it figures it out.

src/app/contexts/layout.tsx
```diff
@@ -0,0 +1,59 @@
...
+            if (newState === undefined) {
+                return defaultLayoutParameters;
+            }
+            if (state.name === newState.name && (
+                JSON.stringify(state.data) === JSON.stringify(newState.data) ||
+                state.name === 'default'
+            )) {
+                return state;
+            }
+            return newState;
```
TomWoodward, line 31: why do we have special handling for the 'default' layout here? shouldn't it be:
```suggestion
            if (state.name === newState.name && JSON.stringify(state.data) === JSON.stringify(newState.data)) {
                return state;
            }
            return newState;
```

also JSON.stringify will give you inconsistent results sometimes, [ts-utils has a utility](https://github.com/openstax/typescript-service-template/blob/main/packages/utils/src/misc/hashValue.ts) that handles hashing objects consistently, [deep-equal](https://www.npmjs.com/package/deep-equal) is also a very small dependency you could use here (that we are already installing anyway)

src/app/contexts/layout.tsx
```diff
@@ -0,0 +1,59 @@
...
+        },
+        defaultLayoutParameters
+    );
+    const layoutData = React.useMemo(
+        () => layoutParameters.name === 'default' ? undefined : layoutParameters.data,
```
TomWoodward, line 36: we shouldn't be enforcing any assumptions in this code about which layouts do or do not accept data, just send the data if it is set in the layoutParameters

src/app/components/shell/router.js
```diff
@@ -31,41 +32,51 @@ const Error404 = loadable({
...
+    const {setLayoutParameters} = useLayoutContext();
+
+    return React.useMemo(
+        () => loadable({
+            loader: () => import(`~/pages/${name}/${name}`),
+            loading,
+            render(loaded, props) {
+                const Component = loaded.default;
+
+                setLayoutParameters();
```
TomWoodward, line 69: calling setState in a render function

src/app/contexts/layout.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+const defaultLayoutParameters: LayoutParameters = {name: 'default'};
+
+function useContextValue() {
+    const [layoutParameters, setLayoutParameters] = React.useReducer(
+        (state: LayoutParameters, newState: LayoutParameters) => {
+            if (newState === undefined) {
+                return defaultLayoutParameters;
+            }
+            if (state.name === newState.name && deepEqual(state.data, newState.data)) {
```
TomWoodward, line 26: suppose at this point you could just do `deepEqual(state, newState)`

TomWoodward (changes requested):

TomWoodward (approved):

PR #2659, opened 2024-08-07
Increase code coverage 6

RoyEJohnson:
[DISCO-448]
This gets to 100% coverage for all TS source files. Sorry it's such a big PR.

[DISCO-448]: https://openstax.atlassian.net/browse/DISCO-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/src/pages/details/book-details-context.js
```diff
@@ -15,8 +23,10 @@ function BookDetailsWithContext({data, children}) {
 }
 
 export default function BookDetailsLoader({slug, children}) {
+    const absoluteSlug = slug.startsWith('/') ? slug : `/${slug}`;
```
jivey, line 26: I can't tell, does this branch get tested by the dual-view test? I don't see an expect for it

jivey (approved):
I'm not familiar with a lot of this, but as it almost exclusively adds types and tests without changing functionality, things seem good.

test/src/pages/details/book-details-context.js
RoyEJohnson: That's just a test helper function, but it doesn't look like any testing code actually does come in with a slug beginning with a slash, so it's an unneeded conditional.

src/app/pages/details/common/links-to-translations.tsx
```diff
@@ -45,14 +45,11 @@ function AnotherLanguage({
...
-    if (!translation) {
-        return null;
-    }
-
     // translation is guaranteed to have a valid value, because the locale
     // is pulled from translations
     return (
-        <a href={`/details/books/${translation.slug}`}>
+         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+         <a href={`/details/books/${translation!.slug}`}>
```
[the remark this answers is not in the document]
RoyEJohnson: The issue is that TS knows that `find` might return `undefined`. But it won't (as the comment notes).
I can use a `?` instead of a `!` if it's less alarming. It doesn't require the eslint-disable that way.

PR #2660, opened 2024-08-08
Add user_id as a hidden field populated from search params

RoyEJohnson:
[DISCO-318]

[DISCO-318]: https://openstax.atlassian.net/browse/DISCO-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/contact/form.js
+++ b/src/app/pages/contact/form.js
@@ -59,6 +59,22 @@ function LabeledInputWithInvalidMessage({
     );
 }
 
+function useAfterSubmit() {
+    const navigate = useNavigate();
+    const {pathname} = useLocation();
+
+    return React.useCallback(
+        () => {
+            if (pathname.includes('embedded')) {
+                window.parent.postMessage('contact form submitted');
+            } else {
+                navigate('/confirmation/contact');
+            }
+        },
+        [navigate, pathname]
+    );
+}
+
 // This is an interim site; normally we can leave postTo null and the default
 // in the salesforceForm will be right.
 const newPostSite = 'https://hooks.zapier.com/hooks/catch/175480/3n62dhe/';
@@ -78,8 +94,6 @@ export default function ContactForm() {
         () => (subject === 'OpenStax Polska') ? '/apps/cms/api/mail/send_mail' : newPostSite,
         [subject]
     );
-    const navigate = useNavigate();
-    const {pathname} = useLocation();
     const onChangeSubject = React.useCallback(
         (value) => setSubject(value),
         []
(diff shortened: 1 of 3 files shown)


src/app/pages/contact/form.js
```diff
@@ -119,7 +124,7 @@ export default function ContactForm() {
             </LabeledInputWithInvalidMessage>
             <LabeledInputWithInvalidMessage className="auto-height" showMessage={showInvalidMessages}>
                 Your Message
-                <textarea cols="50" name="description" rows="6" required />
+                <textarea cols="50" name="description" rows="6" required ref={textAreaRef} />
```
TomWoodward, line 127: ref unnecessary now right?

TomWoodward (changes requested):

src/app/pages/contact/form.js
```diff
@@ -90,11 +91,15 @@ export default function ContactForm() {
...
         [navigate]
     );
+    const searchParams = new window.URLSearchParams(window.location.search);
+    const bodyParams = searchParams.getAll('body').join('\n');
 
     return (
         <SalesforceForm postTo={postTo} afterSubmit={afterSubmit}>
             <input type="hidden" name="external" value="1" />
             <input type="hidden" name="product" value={product} />
+            <input type="hidden" name="user_id" value={searchParams.get('user_id')} />
```
mwvolo, line 101: can we populate this for osweb submissions as well? from `api/user`

RoyEJohnson: So have the form pull it if it's not supplied as a parameter?

TomWoodward: the cookie user is loaded already right? or no? i think we put the cookied user id in here if no user id is passed in the url

RoyEJohnson: I think user cookie is maintained in accounts, not directly visible to osweb.
osweb uses the accounts API to populate a context for user info. That's what I've got it using.

TomWoodward: i think you're understanding what i'm saying and i'm just saying it in a confusing way.

the cookied user is potentially different from the active user in assignable, because assignable uses a session managed by token passing based on the lti launch. so assignable will pass a user uuid to this form and it might be different from the "authenticated user" as far as osweb is concerned, and that is fine. we should use the passed id as a priority and fall back on the user in osweb's standard authentication as a fallback

TomWoodward (approved):

PR #2673, opened 2024-10-01
Fix postMessage in embedded Contact form

RoyEJohnson:
[CORE-532]
This makes the embedded Contact form (mostly) like the [designs](https://www.figma.com/design/9xbDQWGZuE4vdYb8kqsyvj/instructor-side?node-id=3890-4909&node-type=frame&t=FXsvX78xf0ULEwd7-0).

- Display just the form, not the header or mailing address sections
- After form submit, show a minimalist thank you page
- Dismissing the thank you sends a postmessage to the parent window
- Default question topic to OpenStax Assignable for the embedded form

[CORE-532]: https://openstax.atlassian.net/browse/CORE-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/contact/embedded.js
+++ b/src/app/pages/contact/embedded.js
@@ -14,11 +14,12 @@ export default function EmbeddedForm() {
         <main id="maincontent" className="embedded-contact">
             <h1>Report an issue</h1>
             <div>
-                Need to suggest a content correction instead? Fill out the{' '}
-                <a href="/errata" target="_blank">
-                    errata form
-                </a>
-                .
+                Need to suggest a content correction instead? Find the book
+                on our{' '}
+                <a href="/subjects" target="_blank">
+                    subjects page
+                </a>.
+                {' '}You can submit errata from the details page for the book.
             </div>
             <Form setSubmitted={setSubmitted} />
         </main>
--- a/src/app/pages/contact/embedded.js
+++ b/src/app/pages/contact/embedded.js
@@ -1,0 +1,43 @@
+import React from 'react';
+import Form from './form';
+import './embedded.scss';
+
+// File is js because it is dynamically imported (and minimal)
+export default function EmbeddedForm() {
+    const [submitted, setSubmitted] = React.useState(false);
+
+    if (submitted) {
+        return <ThankYou />;
+    }
+
+    return (
+        <main id="maincontent" className="embedded-contact">
+            <h1>Report an issue</h1>
+            <div>
+                Need to suggest a content correction instead? Fill out the{' '}
+                <a href="/errata" target="_blank">
+                    errata form
+                </a>
+                .
+            </div>
+            <Form setSubmitted={setSubmitted} />
+        </main>
+    );
+}
+
+function ThankYou() {
+    const signalDone = React.useCallback(
+        () => window.parent.postMessage('contact form submitted', '*'),
+        []
+    );
+
+    return (
+        <main id="maincontent" className="embedded-contact">
+            <h1>Thanks for contacting us.</h1>
+            <div>Someone from our team will follow up with you soon.</div>
+            <button type="button" className="btn primary" onClick={signalDone}>
+                Done
+            </button>
+        </main>
+    );
+}
--- a/src/app/pages/contact/embedded.js
+++ b/src/app/pages/contact/embedded.js
@@ -27,7 +27,7 @@ export default function EmbeddedForm() {
 
 function ThankYou() {
     const signalDone = React.useCallback(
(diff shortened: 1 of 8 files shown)


src/app/pages/contact/embedded.js
```diff
@@ -0,0 +1,43 @@
...
+    if (submitted) {
+        return <ThankYou />;
+    }
+
+    return (
+        <main id="maincontent" className="embedded-contact">
+            <h1>Report an issue</h1>
+            <div>
+                Need to suggest a content correction instead? Fill out the{' '}
+                <a href="/errata" target="_blank">
```
jivey, line 18: This is outside the scope of these changes but just wondering if we have a card to track this clipping issue on the prod errata page:
image

src/app/pages/contact/embedded.js
```diff
@@ -0,0 +1,43 @@
...
+                .
+            </div>
+            <Form setSubmitted={setSubmitted} />
+        </main>
+    );
+}
+
+function ThankYou() {
+    const signalDone = React.useCallback(
+        () => window.parent.postMessage('contact form submitted', '*'),
```
jivey, line 30: We use screaming case snake for these in assignable, so it would be nice if they all match.

src/app/pages/contact/form.js
```diff
@@ -115,33 +138,48 @@ export default function ContactForm() {
...
-                Your Message
+                <div className="label-text">
+                    Your Message
+                </div>
                 <textarea cols="50" name="description" rows="6" required />
             </LabeledInputWithInvalidMessage>
+            <div className="label-text">
+                Please add a screenshot or any other file that helps explain your request.
+            </div>
             <FileButton name="attachment" />
```
jivey, line 181: Can we add some horizontal padding to the `file-name` text?
image

src/app/pages/contact/embedded.scss
```diff
@@ -0,0 +1,30 @@
+@import 'pattern-library/core/pattern-library/headers';
+
+.embedded-contact {
```
jivey, line 3: We probably need some padding to push the text over or down to accommodate the close button:
image

(and I think this is on your assignments PR, but seems like some padding issue with that button?)

jivey (changes requested):
Mostly small CSS concerns.

src/app/pages/contact/embedded.js
RoyEJohnson: This is the errata page with no book in the URL. With a book, the header gets populated and the selectors at the top are pushed down. Without a book, the page is useless.
https://openstax.org/errata/?book=Principles%20of%20Accounting%2C%20Volume%201%3A%20Financial%20Accounting

src/app/pages/contact/embedded.js
RoyEJohnson: So `CONTACT_FORM_SUBMITTED`, then. I'll need to change it in both PRs.

src/app/pages/contact/form.js
RoyEJohnson: I standardized the styling on this with the contact form.

src/app/pages/contact/embedded.scss
RoyEJohnson: I have reworked the put-away as a back button in a footer (like the design).
PR for that is:
https://github.com/openstax/assignments/pull/192

src/app/pages/contact/embedded.js
jivey: Do we have the book info inside this form to make the right url? Otherwise it seems like we should remove the link.

src/app/pages/contact/form.js
jivey: So fixing it there would fix it for both?

jivey: Oh nevermind I wasn't looking at my updated local os-web instance

src/app/pages/contact/embedded.scss
jivey: It might need some `position: fixed` rules? But I'll let ya'll work that out on that PR 
image

jivey (commented):
Changes are looking good, just have a question about the errata url

src/app/pages/contact/embedded.js
RoyEJohnson: D'oh!

jivey (approved):
The form is looking good, I'm still seeing some layout issues but those are on the assignments side.
image

PR #2674, opened 2024-10-11
Port subjects page to ts

RoyEJohnson:
[CORE-537]
This covers the all-subjects page. The specific-subjects page wll be a separate PR.

[CORE-537]: https://openstax.atlassian.net/browse/CORE-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2675, opened 2024-10-16
Port specific subjects page to ts

RoyEJohnson:
[CORE-537]
This is based on #2674, which should be merged before reviewing this. **Update**: that is done.
Overwhelmingly, this is just adding types and making them work, then writing tests for the code.
In some files, I did some rearranging of code blocks, and reviewing might be best looking at old file and new file side-by-side in separate windows rather than the changes.

[CORE-537]: https://openstax.atlassian.net/browse/CORE-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/carousel/carousel.js
```diff
@@ -41,7 +41,8 @@ function HoverText({which, thing}) {
 export default function Carousel({
     atATime = 1,
     children,
-    hoverTextThing
+    hoverTextThing,
+    ref
```
RoyEJohnson, line 45: Bug fix: had been passing a ref to this, but not using it.

jivey (approved):
Some of the type additions are hard to see because of the file deletions but nothing is jumping out to me as problematic.

PR #2681, opened 2024-11-07
Port details page 3

RoyEJohnson:
[CORE-601]

[CORE-601]: https://openstax.atlassian.net/browse/CORE-601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/pages/details/common/publication-info.tsx
```diff
@@ -19,38 +23,42 @@ function PdfUpdateInfo({updateDate, url}) {
                     />
                 </h4>
                 {formatDate(updateDate)}
-                {
-                    url &&
-                        <div className="callout">
-                            <div>
+                {url && (
```
RoyEJohnson, line 26: This is just a formatting change by prettier

src/app/pages/details/common/resource-box/resource-box-utils.tsx
```diff
@@ -32,7 +70,7 @@ function resourceBoxPermissions({resourceData, resourceStatus, loginUrl}) {
         locked: {
             iconType: 'lock',
             link: {
-                text: <FormattedMessage id='resources.loginToUnlock' />,
+                text: <FormattedMessage id="resources.loginToUnlock" /> as unknown as string,
```
RoyEJohnson, line 73: This feels a little dirty. I'd rather have a non-hook function to generate a string, but I don't find it in the package.

jivey: Is this only used/displayed in `left-content.tsx`? What if you made the text value here the message id, then do something like `<FormattedMessage id={model.link.text} />` in that element, would that work?

test/src/pages/details/desktop-view/instructor-resource-tab.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+            }
+        });
+    });
+    it('returns null until user status exists', async () => {
+        mockUseUserContext.mockReturnValue({});
+        render(<Component />);
+        let caught = false;
+
+        try {
+            await screen.findByRole('heading');
```
jivey, line 59: Maybe you can use`queryByRole` on these, it returns null for a miss. I'm always worried about try/catch swallowing other errors.

RoyEJohnson: I use `find` because there's some promise-resolving that happens. For example, in the next test, I expect a `heading`, but have to wait for it. Here, I wait to be sure it's not showing up.

src/app/pages/details/common/resource-box/resource-box-utils.tsx
RoyEJohnson: The `unlocked` entry also has `link.text` that doesn't use intl, so I'd need to make the recipient handle either `text` or `id` depending on what it's given.

RoyEJohnson: I made a LinkText component to handle it.

test/src/pages/details/desktop-view/instructor-resource-tab.test.tsx
jivey: Oh interesting, maybe then `await expect(screen.findByRole('heading')).rejects.toThrow()` would work, it kinda feels more idiomatic to jest?

src/app/pages/details/common/resource-box/resource-box-utils.tsx
jivey: Nice, I like that

jivey (approved):

test/src/pages/details/desktop-view/instructor-resource-tab.test.tsx
RoyEJohnson: Much nicer, yes. I've replaced the try/catch construct throughout my tests now.

PR #2688, opened 2024-12-05
Add startups sidebar to partners

RoyEJohnson:
[CORE-609]
I have added a phone-format layout. On screens larger than phones, it has the sidebar which has a contrasting heading and a single column of partners. On small screens, the Startups section becomes just another grid at the bottom of the page (phone screens generally hold 2 partner info cards on each row).

[CORE-609]: https://openstax.atlassian.net/browse/CORE-609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/layouts/default/header/menus/main-menu/main-menu.js
+++ b/src/app/layouts/default/header/menus/main-menu/main-menu.js
@@ -15,6 +15,9 @@ import {treatSpaceOrEnterAsClick} from '~/helpers/events';
 import './main-menu.scss';
 
 function DropdownOrMenuItem({item}) {
+    if (! item.name && ! item.label) {
+        return null;
+    }
     if ('menu' in item) {
         return (
             <Dropdown label={item.name} navAnalytics={`Main Menu (${item.name})`}>
--- a/src/app/layouts/default/header/menus/main-menu/main-menu.js
+++ b/src/app/layouts/default/header/menus/main-menu/main-menu.js
@@ -7,7 +7,7 @@ import {
 } from '~/components/language-selector/language-selector';
 import {FormattedMessage} from 'react-intl';
 import {useLocation} from 'react-router-dom';
-import { useDataFromSlug } from '~/helpers/page-data-utils';
+import {useDataFromSlug} from '~/helpers/page-data-utils';
 import Dropdown, {MenuItem} from './dropdown/dropdown';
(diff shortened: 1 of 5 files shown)


src/app/pages/partners/results/results.tsx
```diff
@@ -202,13 +200,51 @@ function resultEntry(pd: PartnerData) {
         cost: pd.affordability_cost,
         rating: pd.average_rating.rating__avg,
         ratingCount: pd.rating_count,
-        partnershipLevel: pd.partnership_level,
+        partnershipLevel: pd.partnership_level as string,
```
jivey, line 203: I think it's already typed as string here https://github.com/openstax/os-webview/blob/add-startups-sidebar-to-partners/src/app/pages/partners/results/results.tsx#L52

jivey (changes requested):
We might need some css tweaks here... I don't think the sidebar card is responsive like the other cards, so it feels a bit disjointed when it is close to other cards (probably needs a margin too):
 image

Also may the filters to vertically stack or allow overflow below 350ish:
image

I'm also wondering about the placement, when you have more filters active, it can feel like it's floating out there... 
image

src/app/pages/partners/results/results.tsx
RoyEJohnson: I made startups display as a normal grid-group if it's there is only one other grid-group.
I think I've got the spacing all fixed up.
The "as string" was leftover from when I had it defined as string | null. Good catch, thanks.
I removed some padding on the mobile-controls row, which was enough to make it good down to 320px screen width.

jivey: Yeah this is a lot better, nice idea to switch up the display based on the available groups.

src/app/layouts/default/header/menus/main-menu/main-menu.js
```diff
@@ -15,6 +15,9 @@ import {treatSpaceOrEnterAsClick} from '~/helpers/events';
 import './main-menu.scss';
 
 function DropdownOrMenuItem({item}) {
+    if (! item.name && ! item.label) {
```
jivey, line 18: Some spaces here?

jivey (approved):

PR #2695, opened 2025-01-14
Strip extra levels off page url

RoyEJohnson:
[CORE-708]

[CORE-708]: https://openstax.atlassian.net/browse/CORE-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/shell/router.js
+++ b/src/app/components/shell/router.js
@@ -126,16 +126,6 @@ function RedirectToCanonicalDetailsPage() {
     );
 }
 
-function RedirectToPageWithoutJunk() {
-    const {name, '*': junk} = useParams();
-
-    if (junk) {
-        return <Navigate to={`/${name}`} replace />;
-    }
-
-    return <TopLevelPage />;
-}
-
 function MainRoutes() {
     const {Layout} = useLayoutContext();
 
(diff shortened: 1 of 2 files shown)


src/app/components/shell/router.js
```diff
@@ -146,7 +156,7 @@ function MainRoutes() {
                 <Route path="/details/" element={<Navigate to="/subjects" replace />} />
                 <Route path="/books/:title" element={<RedirectToCanonicalDetailsPage />} />
                 <Route path="/textbooks/:title" element={<RedirectToCanonicalDetailsPage />} />
-                <Route path="/subjects-preview/*" element={<ImportedPage name="subjects" />} />
+                <Route path="/subjects/*" element={<ImportedPage name="subjects" />} />
```
TomWoodward, line 159: the subjects route exists already somewhere doesn't it? i don't see it but it must...

src/app/components/shell/router.js
```diff
@@ -126,6 +126,16 @@ function RedirectToCanonicalDetailsPage() {
     );
 }
 
+function RedirectToPageWithoutJunk() {
+    const {name, '*': junk} = useParams();
```
TomWoodward, line 130: this ditches everything other than the top url section? so `/foo/bar` will be redirected to `/foo` ? should we not just 404?

TomWoodward (changes requested):

src/app/components/shell/router.js
RoyEJohnson: It was using the `/:name/*` path below, but since that's changing, it needs its own entry.

src/app/components/shell/router.js
RoyEJohnson: We can, but I don't know why or how often we're getting URLs like `/impact/openstax-has-been-a-great-teacher`. I would think it's a better user experience to go to the obvious page.

RoyEJohnson: I have added a commit that 404s instead.

PR #2703, opened 2025-02-06
CORE-671 CookieYes

Dantemss:
- Removed Osano
- Added CookieYes

Tom wants to move this snippet into GTM, so he requested that it be able to run without app code.


Dantemss:
Tested on dev.openstax.org and it seems to work just fine.

src/index.html
```diff
@@ -24,8 +24,78 @@
         <link rel="preconnect" href="https://connect.facebook.net" crossorigin="anonymous">
         <link rel="prefetch" href="<%= process.env.API_ORIGIN %>/apps/cms/api/books/?format=json">
         <link rel="prefetch" href="<%= process.env.API_ORIGIN %>/apps/cms/api/footer/?format=json">
-        <link rel="stylesheet" href="https://ram.openstax.org/osano/osano.css">
-        <script src="https://cmp.osano.com/AzZqbXTbzhHsU3cv1/68d8e8ae-4024-4489-a000-72766ad284a6/osano.js"></script>
+        <script>
```
RoyEJohnson, line 27: Is it possible to move this to a file and have it serve properly?

Dantemss: Since the goal is to move it to GTM soon and remove it from this repo, I think it's not worth moving it? We might not even release it like this.

src/index.html
```diff
@@ -24,8 +24,78 @@
...
+          fetch('/accounts/api/user', { credentials: 'include' }).then((userResponse) => {
```
TomWoodward, line 28: osweb also fetches the user, is this gonna make it fetch the user twice on every page load?

TomWoodward (changes requested):

src/index.html
```diff
@@ -24,8 +24,78 @@
...
+                  headers: { 'Content-Type': 'application/json' },
+                  method: 'PUT',
+                }).then(
+                  (response) => {
+                    if (response.status === 200) {
+                      window.location.reload();
+                    } else {
+                      Sentry.captureMessage(`Failed to save cookie consent (HTTP status code: ${response.status})`);
+                    }
+                  }, (error) => Sentry.captureException(error)
```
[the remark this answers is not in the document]
Dantemss: Good point, I guess I should add a guard and ignore the errors otherwise? I think Sentry is not guaranteed to be accessible when GTM scripts run either.

src/index.html
Dantemss: Yes... I can have both this and the app code write the user to the window (only if not already present), I suppose

src/app/layouts/default/footer/footer.js
```diff
@@ -127,27 +127,11 @@ function Footer({
...
-        [osano]
-    );
-
-    if (osano === undefined || osano.cm.mode === 'debug') {
-        document.body.classList.remove('hide-osano');
-        return null;
-    }
-
-    document.body.classList.add('hide-osano');
+function CookieYesToggle() {
```
Dantemss, line 130: I should move this to ui-components and use it in Rex

TomWoodward: It wasn't done in a shared way before because Roy didn't want to add styled components to osweb

You could pretty much copy/paste though. Maybe rex and assignable could share a component

Dantemss:
This should be ready for another look

src/app/models/accounts-model.ts
```diff
@@ -81,35 +89,27 @@ export default {
...
-    })
+
+        return window._OX_USER_PROMISE.then((user) => {
+            window.dataLayer ||= [];
+            window.dataLayer.push({
+                faculty_status: user.faculty_status
+            });
+            return user;
+        });
+    }, () => { window._OX_USER_PROMISE = undefined; })
```
RoyEJohnson, line 114:
```suggestion
    }, () => {window._OX_USER_PROMISE = undefined;})
```

src/app/models/accounts-model.ts
```diff
@@ -81,35 +89,27 @@ export default {
...
                 console.warn('"Error fetching user info"');
                 return {err};
             }
         );
-    })
+
+        return window._OX_USER_PROMISE.then((user) => {
+            window.dataLayer ||= [];
+            window.dataLayer.push({
+                faculty_status: user.faculty_status
```
RoyEJohnson, line 110:
```suggestion
                faculty_status: user.faculty_status // eslint-disable-line camelcase```

**Review by RoyEJohnson — APPROVED** [review]
A couple of lint errors (it's not part of the check in checks). I have suggested the fixes.

File: src/index.html [hunk 7]
```diff
@@ -24,8 +24,91 @@
...
+          window._OX_USER_PROMISE ||= fetch('/accounts/api/user?always_200=true', {credentials: 'include'}).then(
+              (response) => response.json().catch(
+                  (err) => {
+                      console.warn('No JSON in Accounts response');
+                      return {err};
+                  }
+              ),
+              (err) => {
+                  console.warn('"Error fetching user info"');
+                  return {err};
```
TomWoodward, line 38: doesn't this allow the `{err}` object to proceed to the success handler

Dantemss: Moved around as we discussed

PR #2707, opened 2025-02-19
Use school dropdown on thank you note

RoyEJohnson:
[CORE-756]


src/app/components/form-radiogroup/form-radiogroup.js
```diff
@@ -20,7 +20,7 @@ function Option({item, name, required, selectedValue, onChange}) {
 }
 
 export default function FormRadioGroup({
-    label, longLabel, name, options, required
+    label=undefined, longLabel, name, options, required
```
jivey, line 23: I'm not sure what this is doing here?

jivey (approved):

RoyEJohnson:
I didn't have or want to make a .d.ts file for that JS file. The TS that
imports it assumes everything is a required argument unless it has an
explicit default value.

PR #2712, opened 2025-02-27
Adjust partner page styling

RoyEJohnson:
[CORE-790]

[CORE-790]: https://openstax.atlassian.net/browse/CORE-790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/partners/results/results.scss
+++ b/src/app/pages/partners/results/results.scss
@@ -31,11 +31,39 @@
         border: thin solid ui-color(black);
     }
 
+    /*
+        We want to make breakpoints for each possible number of cards
+        on a row, from 2 to 5, such that the row with a sidebar has one
+        fewer cards per row.
+        first: 2 cards on top, 1 on bottom
+        second: 3/2
+        third: 4/3
+        above: 5/4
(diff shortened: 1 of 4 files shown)


src/app/pages/partners/results/results.scss
```diff
@@ -186,11 +217,8 @@
                 }
 
                 h2 {
-                    @include set-font(h3);
-
-                    background-color: os-color(gray);
-                    color: ui-color(white);
-                    padding: $normal-margin;
+                    // @include set-font(h3);
```
bethshook, line 220: remove?

bethshook (approved):
I'm a little worried about the potential for something breaking here but given all of the constraints I'm not sure there's a better alternative (as you note on the card)

RoyEJohnson:
I made an additional update to make the card-counts explicit in both sections, so that it's not subject to the browser picking slightly different breakpoints (which I saw in Firefox). It makes the CSS simpler, too.

PR #2723, opened 2025-04-01
Port components to ts 4

RoyEJohnson:
[CORE-857]

[CORE-857]: https://openstax.atlassian.net/browse/CORE-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/language-selector/language-selector.tsx
```diff
@@ -19,29 +26,24 @@ export function LanguageLink({locale, slug}) {
...
-}
-
-function LanguageTextUsingIntl({locale}) {
+export function LanguageText({locale}: {locale: string}) {
     const {language} = useLanguageContext();
-    const languageName = React.useMemo(
-        () => new Intl.DisplayNames([language], {type: 'language'}),
-        [language]
+    const text = React.useMemo(
+        () => (new Intl.DisplayNames([language], {type: 'language'})).of(locale),
```
RoyEJohnson, line 32: We don't support browsers so old that Intl.DisplayNames isn't supported, so I took the fallback out.

src/app/components/more-fewer/more-fewer.tsx
```diff
@@ -41,7 +49,7 @@ export function Paginated({children, perPage=10}) {
     }
     useEffect(() => {
         if (pageChanged) {
-            $.scrollTo(ref.current, 70);
+            $.scrollTo(ref.current as HTMLDivElement, 70);
```
jivey, line 52: Don't you want to check for current being set before calling this?  I checked the helper and it doesn't check for a null param so it seems like this could cause issues...

RoyEJohnson: As I understand it, when the `useEffect` runs, the ref will be set, since it's set in the return value for the component. If I do a test for not-set, I won't be able to cover that branch with testing.

RoyEJohnson: I added an `assertNotNull` helper function.

jivey (approved):

PR #2724, opened 2025-04-10
Port components to ts 5

RoyEJohnson:
[CORE-870]
This is based on #2723 which should be merged before reviewing this.

[CORE-870]: https://openstax.atlassian.net/browse/CORE-870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/pages/blog/article/article.tsx
```diff
@@ -220,7 +220,6 @@ function ShareButtons() {
             message={encodeURIComponent(
                 'Check out this OpenStax blog article!'
             )}
-            minimal={true}
```
RoyEJohnson, line 223: The share component is only being used in in this place, which used the `minimal` option, so I made it not an option.

src/app/components/share/share.scss
```diff
@@ -6,31 +6,32 @@ $linkedin-blue: #0077b5;
 
 .share-buttons {
     .buttons {
-        display: flex;
+        display: grid;
+        grid-gap: 1rem;
         justify-content: center;
-        margin: 1rem 0;
-
-        &:not(.minimal) {
```
RoyEJohnson, line 13: The `minimal` option is the only way it was being called, so I got rid of the non-minimal configuration

src/app/components/select/drop-down/drop-down.tsx
```diff
@@ -19,37 +24,39 @@ function RenderItem({item, current, onMouseEnter, onClick}) {
             onMouseEnter={onMouseEnter}
             onClick={onClick}
         >
-            {item.label || item.text}
```
RoyEJohnson, line 22: Everything that uses it now has `label`, not `text`.

src/app/components/salesforce-form/salesforce-form.tsx
```diff
@@ -1,35 +1,21 @@
...
-            }
-            <input type="hidden" name="lead_source" value={leadSource} />
-        </React.Fragment>
-    );
-}
-
-function SfForm({children, postTo, afterSubmit}) {
+function SfForm({children, postTo, afterSubmit}: FormParams) {
     const [listening, setListening] = React.useState(false);
-    const {webtocaseUrl, debug, oid} = useSalesforceContext();
```
RoyEJohnson: Nothing uses webtocaseUrl anymore.

src/app/components/body-units/quote.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
RoyEJohnson, line 1: This was moved from being its own component to being part of body-units, since nothing else used it. Also simplified it a bit, since there was an image feature that wasn't being used.

src/app/components/progress-ring/progress-ring.tsx
```diff
@@ -1,17 +1,12 @@
 import React from 'react';
 import './progress-ring.scss';
 
-function Circle({basicProps, className, strokeDashoffset=0}) {
```
RoyEJohnson: I wasn't abstracting anything! It was a useless layer of component.

test/src/components/paginator.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {PaginatorControls} from '~/components/paginator/search-results/paginator';
+import SimplePaginator from '~/components/paginator/simple-paginator';
+
+function activePage() {
+    const activeButton = screen.getByRole('button', {current: 'page'});
+
+    return activeButton.textContent;
+}
+
+describe('paginator', () => {
```
RoyEJohnson, line 14: Didn't have a describe wrapper on it before, so it updates all the lines. Only had one test before; now two, so not a big file anyway.

src/app/pages/blog/gated-content-dialog/gated-content-dialog.tsx
```diff
@@ -73,16 +73,13 @@ function GatedContentBody({
 function SubjectSelector() {
     const {subjectSnippet: data} = useBlogContext();
     const options = React.useMemo(
-        () =>
-            camelCaseKeys(
```
jivey, line 77: Is this transformation happening somewhere else now?

src/app/components/share/facebook-script.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
+(function(d, s, id) {
+  var js, fjs = d.getElementsByTagName(s)[0];
+  if (d.getElementById(id) || !fjs) return;
+  js = d.createElement(s); js.id = id;
+  js.src = '//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.8';
+  fjs.parentNode.insertBefore(js, fjs);
+}(document, 'script', 'facebook-jssdk'));
+/* eslint-enable */
```
jivey, line 9: I don't see this being used in the PR?

jivey (commented):
Seems good just some small questions

src/app/pages/blog/gated-content-dialog/gated-content-dialog.tsx
RoyEJohnson: I realized that camelCaseKeys has no effect, since the keys are all known here (`label` and `value`)

src/app/components/share/facebook-script.js
RoyEJohnson: It's [imported](https://github.com/openstax/os-webview/blob/port-components-to-ts-5/src/app/components/share/share.tsx#L7) by the "share" component.

PR #2726, opened 2025-04-17
Core 898 landing flex footer

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-898

Add a new footer for landing layout flex pages.

I am a bit new to this code, but it seemed like this combination of `import` and `JITLoad` was a common way to avoid importing styles/code that isn't actively used. I realize that the footer loader has its own loading placeholder which is then placed inside the `JITLoad` which also has a loading placeholder. Maybe there's a better way that I overlooked.

I am also wondering if (1) the condition for flex pages is correct and (2) if I should export a list of flex page types somewhere that can be used here and in `fallback-to` so that this condition is not duplicated.

## Desktop

Screenshot 2025-04-16 at 3 09 33 PM

## Mobile

Screenshot 2025-04-16 at 3 09 48 PM

## Contact Us Dialog (iframe)

Screenshot 2025-04-17 at 4 47 56 PM

### DIFF
--- a/test/src/layouts/landing/footer.test.tsx
+++ b/test/src/layouts/landing/footer.test.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
-import userEvent, { UserEvent } from '@testing-library/user-event';
+import userEvent from '@testing-library/user-event';
 import {render, screen, waitFor} from '@testing-library/preact';
 import {describe, it} from '@jest/globals';
 import { MemoryRouter } from 'react-router-dom';
 import { useContactDialog } from '~/layouts/landing/footer/flex';
+
+const user = userEvent.setup({advanceTimers: jest.advanceTimersByTime});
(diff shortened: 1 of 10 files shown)


test/src/layouts/landing/footer.test.tsx
```diff
@@ -1,35 +1,39 @@
...
             Contact Us
-            <ContactDialog contactFormParams={contactFormParams} />
+            <ContactDialog {...props} />
         </button>
     );
 }
 
 describe('flex landing footer', () => {
     describe('contact dialog', () => {
+        let user: UserEvent;
```
RoyEJohnson, line 21: It's been my experience that I can set up user once, rather than needing to set it up before each test.

RoyEJohnson (approved):

PR #2736, opened 2025-05-21
add book list component for flexpages

TomWoodward:
https://openstax.atlassian.net/browse/CORE-900

### DIFF
--- a/src/app/pages/flex-page/blocks/RichTextBlock.scss
+++ b/src/app/pages/flex-page/blocks/RichTextBlock.scss
@@ -9,12 +9,12 @@
     width: 100%;
     height: auto;
   }
-  
+
   img.richtext-image.left {
     float: left;
     margin-right: $normal-margin;
   }
-  
+
   img.richtext-image.right {
     float: right;
     margin-left: $normal-margin;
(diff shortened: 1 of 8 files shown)


src/app/pages/flex-page/flex-page.scss
```diff
@@ -0,0 +1,4 @@
+
+.page.flex-page {
+  overflow: visible;
```
TomWoodward, line 3: the overflow:hidden on the page was hiding the dropdown menu if there was no following content

src/app/pages/flex-page/blocks/RichTextBlock.scss
```diff
@@ -23,7 +23,7 @@
   > *:first-child {
     margin-top: 0;
   }
-  > *:last-child {
+  > *:not(h1, h2, h3, h4, h5, h6):last-child {
```
TomWoodward, line 26: not certain this is the best way to handle this, but without the margin the text descenders are cut off

RoyEJohnson: This appears to be a `line-height` issue in the pattern library. hero-h1 line-height is `5rem`. It should be `normal` rather than being specified. This dates back to ancient times when the pattern-library was first spec'd.

PR #2750, opened 2025-06-25
Core 1039 improve portal routing

RoyEJohnson:
[CORE-1039]
By examining the data from the CMS to determine what kind of page was being requested, I was able to get rid of the fallback process.

[CORE-1039]: https://openstax.atlassian.net/browse/CORE-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/shell/router-helpers/page-routes.tsx
+++ b/src/app/components/shell/router-helpers/page-routes.tsx
@@ -55,6 +55,7 @@ export function DetailsRoutes() {
 
 export function OtherPageRoutes() {
     const dir = assertDefined(useParams().dir);
+    const {'*': path} = useParams();
 
     if (['books', 'textbooks'].includes(dir)) {
         return (
@@ -65,6 +66,10 @@ export function OtherPageRoutes() {
                 />
             </Routes>
         );
+    }
+
+    if (dir === 'general') {
+        return <Navigate to={`/${path}`} replace />;
     }
 
     if (dir === 'home') {
--- a/test/src/components/shell.test.tsx
+++ b/test/src/components/shell.test.tsx
@@ -65,7 +65,7 @@ describe('shell', () => {
             case 'pages/general-page':
                 return camelCaseKeys(transformData(generalPage));
             default:
-                if (path.startsWith('errata/')) {
+                if (path.startsWith('errata/') || path.startsWith('pages/')) {
                     return [];
                 }
                 console.info('*** Requested', path);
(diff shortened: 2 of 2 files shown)


src/app/components/shell/router-helpers/page-routes.tsx
```diff
@@ -0,0 +1,114 @@
...
+    FlexPageData,
+    useDataToSetLayoutParameters
+} from '~/pages/flex-page/flex-page';
+import useLayoutContext from '~/contexts/layout';
+import {assertDefined} from '~/helpers/data';
+import {GeneralPageFromSlug} from '~/pages/general/general';
+import {ImportedPage} from './page-loaders';
+import {RouteAsPortalOrNot} from './portal-page-routes';
+
+type PageData = FlexPageData & {
```
RoyEJohnson, line 14: This file is the bulk of the new routing process.

src/app/components/shell/router.tsx
```diff
@@ -0,0 +1,95 @@
+import React, {useEffect} from 'react';
```
RoyEJohnson, line 1: What was left in router.js is now TS.

src/app/layouts/landing/landing.tsx
```diff
@@ -7,17 +7,17 @@ import useMainClassContext, {
 import useLanguageContext from '~/contexts/language';
 import ReactModal from 'react-modal';
 import cn from 'classnames';
-import { LinkFields } from '../../pages/flex-page/components/Link';
+import {LinkFields} from '../../pages/flex-page/components/Link';
```
RoyEJohnson, line 10: Changes here are just due to prettier

src/app/components/shell/router-helpers/page-routes.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    return <RouteAsPortalOrNot />;
+}
+
+
+// There are a couple of pages whose names in the CMS don't match their osweb urls
+const mismatch: Record<string, string> = {
+    press: 'news',
+    'edtech-partner-program': 'openstax-ally-technology-partner-program'
```
jivey, line 81: Is this working for you? On my end I'm seeing it 404 this page, when before it loaded.

jivey: Yeah if I remove the /general dir then they load

RoyEJohnson: Ahh, I wasn't using the "/general". I think that should be considered obsolete. General pages can be loaded without it. If we do want to have that, I'll need to put a route back in for it; the issue you're seeing is that it's being treated as a portal (the flex page for which doesn't exist).

RoyEJohnson: I can put a redirect in (/general/something -> /something). That's very much the preferred address.

PR #2754, opened 2025-07-09
CORE-1064: Port K12 page to TS

RoyEJohnson:
[CORE-1064]
It looks like these files mostly show as updates rather than delete-and-creates, even in the final (post-prettier) commit.
I renamed all-subjects to k12-main, because it was clearer, so there are a bunch of SCSS files that show as renamed without changes.
contact and blogs components were commented out, so I got rid of them.

[CORE-1064]: https://openstax.atlassian.net/browse/CORE-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/pages/k12/all-subjects/all-subjects.js
```diff
@@ -1,39 +0,0 @@
-import React from 'react';
```
RoyEJohnson, line 1: I changed all-subjects to k12-main, which seemed clearer.

PR #2758, opened 2025-07-24
30 July release and hotfix

RoyEJohnson:
This PR was to fix the contact form as displayed from landing pages. Then we made it the release candidate and subsequently did a little hotfix.

### DIFF
--- a/src/app/components/select/select.tsx
+++ b/src/app/components/select/select.tsx
@@ -27,9 +27,7 @@ function SValidationMessage({
 }) {
     const {item} = useSelectContext();
 
-    return <ValidationMessage
-        watchValue={item} elementRef={elementRef}
-    />;
+    return <ValidationMessage watchValue={item} elementRef={elementRef} />;
 }
 
 export default function Select({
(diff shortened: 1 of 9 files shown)


src/app/components/select/select.tsx
```diff
@@ -27,9 +27,7 @@ function SValidationMessage({
 }) {
```
RoyEJohnson, line 27: These are just reformatting courtesy of "prettier".

src/app/pages/contact/form.tsx
```diff
@@ -95,7 +95,9 @@ function useAfterSubmit(setSubmitted?: SetSubmitted) {
 
 function useSubjectWithInitialization() {
     const isEmbedded = useIsEmbedded();
-    const initialValue = isEmbedded ? 'OpenStax Assignable' : 'General';
+    const searchParams = new window.URLSearchParams(useLocation().search);
```
RoyEJohnson, line 98: This tests to see whether there are multiple body parameters passed in. For the normal embedded form, there will be only one. Assignable passes multiple. Previously, the embedded form was set up to default to Assignable.

PR #2815, opened 2026-01-13
CORE-1439: add missing accessible names

jomcarvajal:
[CORE-1439]

[CORE-1439]: https://openstax.atlassian.net/browse/CORE-1439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/details/common/resource-box/resource-box.tsx
+++ b/src/app/pages/details/common/resource-box/resource-box.tsx
@@ -69,7 +69,7 @@ function Bottom({model}: {model: ResourceModel}) {
         <div className="bottom">
             <LeftContent model={model} />
             {model.printLink && (
-                <a className="print-link" aria-label="Buy print" href={model.printLink}>
+                <a className="print-link" href={model.printLink}>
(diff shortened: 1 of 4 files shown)


src/app/pages/details/common/resource-box/resource-box.tsx
```diff
@@ -69,7 +69,7 @@ function Bottom({model}: {model: ResourceModel}) {
         <div className="bottom">
             <LeftContent model={model} />
             {model.printLink && (
-                <a className="print-link" href={model.printLink}>
+                <a className="print-link" aria-label="Buy print" href={model.printLink}>
```
RoyEJohnson: This is just a link, and it has text in it, so it shouldn't need a label. I think the "buy print" that was mentioned on the card is the nav in order-print-copy.tsx (which you added a label to already).

RoyEJohnson (changes requested):

RoyEJohnson (approved):

PR #2854, opened 2026-04-15
adds new tests for k12 and kinetic links

omehes:
https://openstax.atlassian.net/browse/CORE-988

### DIFF
--- a/e2e_tests/ui/pages/home.py
+++ b/e2e_tests/ui/pages/home.py
@@ -1,6 +1,7 @@
 import pytest
 
 import re
+from playwright.async_api import expect
 
 
 class HomeRex:
@@ -252,7 +253,6 @@ class HomeRex:
 
     # K12 and Kinetic pages
 
-    @pytest.mark.asyncio
     async def click_osweb_k12_link(self):
         k12 = self.page.get_by_role("link", name=re.compile(r"Explore K12"))
         await k12.scroll_into_view_if_needed()
@@ -269,7 +269,11 @@ class HomeRex:
     @pytest.mark.asyncio
     async def click_k12_find_your_subject_book_cards_science_menu(self):
         await self.page.mouse.wheel(0, 1500)
-        await self.page.get_by_role("radio", name="Science").click()
+        radiogroup = self.page.locator('[role="radiogroup"]')
+        await radiogroup.wait_for(state="visible")
+        science_radio = self.page.get_by_role("radio", name="Science")
+        await science_radio.scroll_into_view_if_needed()
+        await science_radio.click()
 
     @pytest.mark.asyncio
     async def k12_find_your_subject_book_card_astronomy_is_visible(self):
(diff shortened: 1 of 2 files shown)


e2e_tests/ui/pages/home.py
```diff
@@ -248,6 +250,43 @@ def free_digital_library_subjects(self):
     def search_technology_partners_link(self):
         return self.page.get_by_role("link", name="Search Technology Partners")
 
+    # K12 and Kinetic pages
+
+    @pytest.mark.asyncio
```
jivey: Do these need this decorator if they aren't tests? (no `test_` prefix?)

e2e_tests/ui/pages/home.py
```diff
@@ -248,6 +250,43 @@ def free_digital_library_subjects(self):
...
+    async def click_osweb_kinetic_link(self):
+        kin_locator = self.page.get_by_role("link", name="Try OpenStax Kinetic")
+        await kin_locator.scroll_into_view_if_needed()
+        await kin_locator.click()
+
+    @pytest.mark.asyncio
+    async def kinetic_page_sample_study_link_is_enabled(self):
+        study_locator = self.page.get_by_role("link", name="Try a sample study")
+        await study_locator.scroll_into_view_if_needed()
+        return await study_locator.is_enabled()
```
jivey: I'm not sure how a link will ever be disabled, so you might be able to remove this.

e2e_tests/ui/pages/home.py
```diff
@@ -248,6 +250,43 @@ def free_digital_library_subjects(self):
...
+    def k12_find_your_subject_dropdown(self):
+        return self.page.locator("div.buttons select.classic")
+
+    @property
+    def k12_find_your_subject_dropdown_options(self):
+        return self.page.locator("div.buttons select.classic option[value]")
+
+    @pytest.mark.asyncio
+    async def click_k12_find_your_subject_book_cards_science_menu(self):
+        await self.page.mouse.wheel(0, 1500)
```
jivey, line 271: I'm worried this will be brittle to layout changes, could we use a scroll into view method instead?

e2e_tests/ui/pages/home.py
omehes: Done

e2e_tests/ui/pages/home.py
omehes: Done

e2e_tests/ui/pages/home.py
omehes: I dont know why, but scroll into view did not work on this element or any around it. That is why I had to find another way...

omehes:
@jivey Thanks for the review. All fixed, ready for another go. The one with mouse scroll, is partially fixed. For some reason, the scroll into view does not work on that element, unless I scroll down some... tried all kinds of things, Claude suggested things but none helped...

jivey (approved):

PR #2855, opened 2026-04-15
adds github actions for QA tests to run at each PR

omehes:


omehes:
@Dantemss I think, the next step here is for you to add user and password. The same as in rex.

Dantemss:
@omehes do we have a dedicated test user for this? I don't remember adding the one for Rex.

Dantemss (approved):

PR #2875, opened 2026-05-26
Core 2238: Restore navigation to 404 page

RoyEJohnson:
[CORE-2238]

[CORE-2238]: https://openstax.atlassian.net/browse/CORE-2238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/pages/404/404.tsx
+++ b/src/app/pages/404/404.tsx
@@ -9,9 +9,11 @@ export default function Page() {
     });
     const {layoutParameters, setLayoutParameters} = useLayoutContext();
 
-    if (layoutParameters.name === null) {
-        setLayoutParameters();
-    }
+    React.useEffect(() => {
+        if (layoutParameters?.name === null) {
+            setLayoutParameters?.();
+        }
+    }, [layoutParameters?.name, setLayoutParameters]);
 
     return (
         <main className="not-found no-style page">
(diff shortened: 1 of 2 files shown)


src/app/pages/404/404.tsx
```diff
@@ -1,11 +1,17 @@
...
+import useLayoutContext from '~/contexts/layout';
 
 export default function Page() {
     useDocumentHead({
         title: '404 Not Found - OpenStax',
         noindex: true
     });
+    const {layoutParameters, setLayoutParameters} = useLayoutContext();
+
+    if (layoutParameters.name === null) {
```
TomWoodward: won't this make loading a 404 page in a portal flash the main nav while the page data is loading?

TomWoodward (changes requested):

src/app/pages/404/404.tsx
RoyEJohnson: Not that I can see. A top-level portal page won't do a 404 because it has to exist in the CMS to be a portal page. A lower-level page within the portal doesn't load until the layout parameters are set, so this if condition isn't met.

TomWoodward (approved):

PR #2878, opened 2026-05-29
Refactor button styles

TomWoodward:
removes global `.btn,button` styles and re-applies them locally where necessary, also keeps them in any raw cms content.

pushes common focus indicator styling down to the pattern-library, and bumps the underlying flex-page-renderer pattern library so everything is all consistent

the global `.btn,button` styles were not actually the same as the button macro ones, i preferred the pattern-library styles and removed the divergent ones so everything is aligned

pattern library pr is https://github.com/openstax/pattern-library/pull/77, this is included here already


src/app/pages/research/research.scss
```diff
@@ -60,6 +72,12 @@
         padding-bottom: 2rem;
     }
 
+    // `.py-4` is a padding utility, but the research "view all" toggles are
+    // `role="button"` spans/paragraphs carrying it — give those a focus ring.
+    .py-4[role="button"] {
```
TomWoodward: i would add another class to select this with instead of relying on the unrelated padding utility

src/app/components/dialog/dialog.scss
```diff
@@ -51,6 +51,8 @@ dialog,
 
     .put-away {
         @include body-font(3.6rem);
+        @include button-reset();
+
```
RoyEJohnson, line 55: The next two lines can be removed, as they're redundant with button-reset now, right?

src/app/pages/partners/active-filters/active-filters.scss
```diff
@@ -43,6 +43,8 @@
             }
 
             button {
+                @include button-reset();
+
```
RoyEJohnson, line 47: again the next two lines could be removed

src/app/pages/assignable/sections/cta/cta.scss
```diff
@@ -29,4 +29,10 @@
     .primary {
         @extend %primary;
     }
+
+    // CTA buttons are large and uppercased.
+    .btn {
+        padding: 1.5rem 2rem;
+        text-transform: uppercase;
```
RoyEJohnson: The "Let us know" button was previously not uppercase. Is this a design change or the Assignable page?

src/app/pages/ally-logos/ally-logos.scss
```diff
@@ -36,6 +36,10 @@
         .url {
             padding: 0.5rem 0 0.5rem 1rem;
         }
+
+        button {
```
RoyEJohnson, line 42: These (/ally-logos, the copy buttons) lost padding and foreground/background color.

RoyEJohnson (changes requested):
There are two places where I noticed design changes.

src/app/components/dialog/dialog.scss
TomWoodward: Done — removed `background-color: transparent` and `border: 0`, both redundant with `button-reset()`. Left `box-shadow: none` and `color: inherit`, since `button-reset()` does not set those. (487a4fc)

src/app/pages/partners/active-filters/active-filters.scss
TomWoodward: Done — dropped `appearance: none` and `background-color: transparent` (redundant with `button-reset()`); kept `color`, `padding`, and `transform`. (487a4fc)

src/app/pages/assignable/sections/cta/cta.scss
TomWoodward: Good catch — it was unintended. The old global rule was `.cta > .btn` (direct child), but this button is a grandchild (`.cta > .content > a.btn`), so it was never uppercased before; the new descendant-scoped `.cta .btn` rule newly applied it. (The `padding: 1.5rem 2rem` in that rule was also inert — `%primary`'s button-basic overrides it to `0 30px`.) Removed the block, so it is back to non-uppercase, matching production. (487a4fc)

src/app/pages/ally-logos/ally-logos.scss
TomWoodward: Fixed — restored the copy button via `@include button-type-specific(secondary)` + `padding: 1rem`, so it is the gray (#757575) fill with a white icon and 1rem padding again. Verified it matches the pre-refactor dev site. (487a4fc)

RoyEJohnson (approved):

PR #2880, opened 2026-06-03
small selector changes

omehes:


e2e_tests/ui/pages/home.py
```diff
@@ -209,11 +209,11 @@ async def click_campus_affordability_link(self):
 
     @property
     def higher_education_bookstore_order_access_code(self):
-        return self.page.get_by_role("link", name="Order access codes")
+        return self.page.get_by_role("link", name="Order Assignable access codes")
 
     @property
     def higher_education_bookstore_view_print_options(self):
-        return self.page.get_by_role("link", name="View print options")
+        return self.page.get_by_role("link", name="Place a bulk print bookstore order")
```
mwvolo, line 216: it's really up to you, and maybe it's a good idea for very important pages, but i would avoid testing the actual content coming from the cms.
checking to see if a button or link that's expected is there seems reasonable, but I suggest focusing on things that could break with code changes, not just an update in the CMS (totally open to other approaches though)

mwvolo (approved):

PR #2881, opened 2026-06-05
Add add-to-calendar button to upcoming webinar tiles

RoyEJohnson:
[CORE-1406]

[CORE-1406]: https://openstax.atlassian.net/browse/CORE-1406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yarn.lock
```diff
@@ -4753,6 +4753,20 @@ acorn@^8.1.0, acorn@^8.11.0, acorn@^8.15.0, acorn@^8.16.0, acorn@^8.2.4, acorn@^
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-8.16.0.tgz#4ce79c89be40afe7afe8f3adb902a1f1ce9ac08a"
   integrity sha512-UVJyE9MttOsBQIDKw1skb9nAwQuR5wuGD3+82K6JgJlm/Y+KI92oNsMNGZCYdDsVtRHSak0pcV5Dno5+4jh9sw==
 
+add-to-calendar-button-react@^2.14.0:
```
jivey, line 4130: At first I was worried about the fetch in this library but that only seems to happen when passing the pro key prop so probably not a concern.

jivey (approved):

PR #2885, opened 2026-06-10
only trigger playwright tests on a push to main

mwvolo:
This pull request updates the Playwright GitHub Actions workflow to trigger on pushes to the `main` branch instead of pull requests.

Workflow trigger update:

* [`.github/workflows/playwright.yml`](diffhunk://#diff-7afcd2d8f7b49bda74843f209eefb7b2da45f7e7803bf2e4bd636699b76aa2d3L3-R5): Changed the workflow trigger from `pull_request` to `push` events on the `main` branch.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,6 +1,8 @@
 name: Playwright
 
-on: [pull_request]
+on:
+  push:
+    branches: [main]
 
 defaults:
   run:


omehes:
@Dantemss Is this what we want with playwright tests running on PRs, or merged to main or...? We had a conversation about this a while ago but we did not come to a solution... I think it could work. rex is set that way. You also mentioned that time something to adjust on deploy part... @mwvolo

mwvolo:
yeah, I was just getting tired of every PR showing it was failing. I'm okay with whatever solution works, I think this is where we landed.
I think this was the thread https://openstax.slack.com/archives/C07K4CU96N6/p1778867657812869

omehes:
yes, that's the conversation. This is what @Dantemss proposed and this might be OK, but I need his help:

`need to change the conditions on playwright.yml to look only at the e2e folder plus allow manual runs (specifying the base url), then change the deployment scripts to trigger the runs after deployment`

omehes (approved):

Dantemss (approved):
I had suggested something more complicated but this is fine as well. Let's see how it does.

mwvolo:
I was thinking about adding something like this:
```
on:
  push:
    branches: [main]
  workflow_dispatch:
    inputs:
      base_url:
        description: 'Base URL to test against'
        required: false
        default: 'https://staging.openstax.org'
```
but i'm not sure how this stuff is working, so probably best for one of you to decide what 'more complicated' looks like

omehes:
I think this can be it, what you re-suggested @mwvolo 
It gives the manual option of adding base url, which could be useful. We can try this out and see. What do you think @Dantemss

mwvolo:
@omehes I updated staging so your tests wouldn't fail on the menu change.

PR #2887, opened 2026-06-10
adds new test to check for broken images

omehes:


omehes:
Fails are due to changes as per Michael

mwvolo:
I reverted the menu on staging, so I think these might be good now

RoyEJohnson (approved):

Issue #1703, opened 2020-03-17
First banner image on homepage is somewhat fuzzy


RoyEJohnson:
# Description
The first image in the homepage banner is lower-resolution than the other banner images and generally looks a little fuzzy.
https://d3bxy9euw4e147.cloudfront.net/oscms-prodcms/media/original_images/os-support-banner-desktop.png

The other images are 1937x968 and in JPG format. The first image is 1600x800 and a PNG. It may have been scaled down from its original size.

# Acceptance criteria
First banner image looks crisp

victoria-cooke:
@allisonhowell 
Larger images attached:
[os-support-banner-desktop-3.png](https://images.zenhubusercontent.com/5c0fedb9d1fab72f0d70aeaa/23157662-e0f2-493c-b7c3-579c46806574)
[os-support-banner-mobile-3.png](https://images.zenhubusercontent.com/5c0fedb9d1fab72f0d70aeaa/015a43f3-07d4-4f22-ad39-a77c076a485b)

PR #2906, opened 2026-06-19
Style flex-page rich-text features (eyebrow, brand colors)

mwvolo:
## What
Styles the CORE-2302 rich-text editor features for flex pages — the `eyebrow` label and the `text-blue/green/orange/gray/white` brand colors — under the fully-qualified `.flex-page.page div.content-block-rich-text` selector.

## Why here
Per Tom's review on openstax/flex-pages#10: these are openstax-cms / Draftail-specific classes, so they belong in the consuming app rather than the generic flex-page renderer. The renderer keeps only generic styling (default heading color, `figcaption`); big number became its own block (flex-pages #10 + openstax-cms #1703).

## Companions
- openstax-cms #1698 registers these Draftail features (already merged).
- flex-pages #10 removed this styling from the renderer.

### DIFF
--- a/src/app/pages/flex-page/flex-page.scss
+++ b/src/app/pages/flex-page/flex-page.scss
@@ -1,4 +1,27 @@
+@import 'pattern-library/core/pattern-library/headers';
 
 .page.flex-page {
   overflow: visible;
+}
+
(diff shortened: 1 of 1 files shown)


RoyEJohnson (approved):
Interesting choice for letter spacing.

PR #2918, opened 2026-06-29
CORE-2399: Remove Collections section from Blog page

RoyEJohnson:
[CORE-2399]

[CORE-2399]: https://openstax.atlassian.net/browse/CORE-2399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/explore-by/explore-by.tsx
+++ b/src/app/components/explore-by/explore-by.tsx
@@ -1,18 +1,18 @@
 import React from 'react';
 import {Link, useLocation} from 'react-router-dom';
-import {ExploreItem, isCategory} from './types';
+import {Category} from './types';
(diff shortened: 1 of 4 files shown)


mwvolo (approved):
LGTM. Does this come with anything we can rip out of the CMS?
I've noticed the bots do a lot better when they aren't trying to figure out what's used and what's not, so I've been trying to clean up legacy stuff in both repos.

RoyEJohnson:
I don't _think_ there's anything in the CMS data that we can remove due to this. Collections still exist and are used, just not in this particular way.

PR #2921, opened 2026-07-01
perf(chat): defer Salesforce chat widget load until browser idle

mwvolo:
## Summary
- Traced the subjects/book-details pages with Chrome DevTools and confirmed the Salesforce Embedded Messaging widget (`chat_subjects`/`chat_book_details` flags, gated to logged-in users) pulls in a ~17-request Lightning Web Runtime chain that fired synchronously the moment the `Chat` component mounted, racing the page's own render/data fetching.
- Defer the bootstrap script load with `requestIdleCallback` (falling back to a 200ms timeout for browsers without it) so it only starts once the browser is idle, and add `rel="preconnect"` hints for the Salesforce origins so the connection is warm by the time it loads. The already-loaded short-circuit path is unchanged (still synchronous).

## Test plan
- [x] `yarn jest test/src/components/chat.test.tsx` — all 16 tests pass (updated to flush the idle-deferred timer before asserting on the script)
- [x] `yarn lint:ts` / `yarn lint:js` — clean
- [x] Deploy to dev and re-run a Chrome DevTools performance trace on `/subjects` and `/details/*` while logged in to confirm the LWR chain no longer fires on initial mount

### DIFF
--- a/src/app/components/chat/chat.tsx
+++ b/src/app/components/chat/chat.tsx
@@ -8,6 +8,8 @@ type VisibleFieldArg = {
     isEditableByEndUser: boolean
 }
 
+const IDLE_CALLBACK_FALLBACK_DELAY = 200;
+
 declare global {
     interface Window {
         embeddedservice_bootstrap?: {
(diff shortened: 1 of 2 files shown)


RoyEJohnson (approved):
Looks sensible.

PR #2924, opened 2026-07-03
v2.160.2 :: hotfix button styles

mwvolo:


mwvolo:
@RoyEJohnson I always forgot what happens to these PRs. This is released on prod.

mwvolo:
are you opposed to using release tags instead of doing it this way? it throws me that the repos are doing releases differently.
I was thinking about setting up quarterly releases for web (so people knew where to put work and when to expect it). What do you think about shifting to using a v2027.q1 type release numbering? I'm not sure if that would mess anything up, but I don't think so.  
Then it would be like 
```
os-webview v2026.q1
openstax-cms v2026.q1
```
if something had to be released more than once a quarter is would be a `v2026.q1.2` (something like this... maybe talking through it with claude would help)

RoyEJohnson:
I don't have a strong attachment to using version numbers as the release tag. I do like having the actual Release entry as well as the tags, because it keeps track of which tags were actually released, when.
The PR typically gets merged after release. Ideally, the associated changes are minimal (updated version number, upgraded packages).

RoyEJohnson:
I'm confused by the seemingly-backward wanting to merge 16 commits into
[release-v2.160.2](https://github.com/openstax/os-webview/tree/release-v2.160.2)
from
[main](https://github.com/openstax/os-webview/tree/main)

mwvolo:
oh, so I basically created this completely wrong lol - sorry about that

I'm going to cut a new release soon, I'll send it to you in Slack so we can see if it might be a good approach. I think the downside is if you have to hotfix, you'll need to create a branch from that tag and apply the change, but I've actually found that easier than cherry picking.

PR #2934, opened 2026-07-13
Fix .text-orange/.text-green Draftail color tokens

mwvolo:
Companion to flex-pages #25.

`.text-orange` `.text-green` (Draftail rich-text classes applied inside flex-page content) were resolving to the illustration-tuned `os-color(orange)` `os-color(green)` swatches. The site's custom-CSS override shows they should read as `ui-color(primary)` (#d4450c) and `os-color(deep-green)` (#0c9372) instead.


mwvolo:
~I'm not sure if this is actually needed, but it's part of the overrides we had on the site to match designs.~
it's very localized to flex page styles, so i think it's fine

PR #2936, opened 2026-07-15
Switch flex-page-renderer to GitHub Packages registry

mwvolo:
## Summary
- flex-pages now publishes to `npm.pkg.github.com` instead of git-tag dist branches (per Tom's new setup, demoed in `assessment-components` PR [#114](https://github.com/openstax/assessment-components/pull/114))
- Adds `.npmrc` mapping `@openstax` scope to the GitHub Packages registry
- CI (`build.yml`) gets `packages: read` permission and an auth step using `GITHUB_TOKEN` before install
- Bumps `@openstax/flex-page-renderer` from the old `renderer-1.1.16` git tag to `^1.1.18`, the version currently published to the registry

## Setup required (one-time, per developer)
```
npm config set -g //npm.pkg.github.com/:_authToken $(gh auth token)
```
Without this, `yarn install` will fail to fetch `@openstax/*` packages.

## Test plan
- [x] `yarn install` resolves `@openstax/flex-page-renderer@1.1.18` from `npm.pkg.github.com`
- [x] `yarn lint:js` / `yarn lint:ts` pass
- [x] `./script/build` succeeds
- [x] `./script/test` — 178 suites / 725 tests pass (pre-existing coverage gap in 3 unrelated files, not caused by this change)
- [ ] Confirm CI's `GITHUB_TOKEN` actually has read access to packages published under the `flex-pages` repo (cross-repo package visibility can be restrictive by default — flag to Tom if the workflow 404s)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,12 +17,12 @@ jobs:
           cache: yarn
       -
         name: "Auth to GitHub Packages"
-        run: echo "//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}" >> ~/.npmrc
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo '//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}' >> ~/.npmrc
       -
         name: "Setup environment"
         run: ./script/setup
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       -
         name: "Lint osweb"
         run: yarn lint
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,9 @@
 jobs:
   osweb:
     runs-on: ubuntu-22.04
+    permissions:
+      contents: read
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -12,6 +15,11 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: yarn
+      -
+        name: "Auth to GitHub Packages"
+        run: echo "//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}" >> ~/.npmrc
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
TomWoodward: i think this'll work, but its not the way the other projects are set up, and it results with the secret written to that file on disk, which may not be ideal.

if you look at https://github.com/openstax/typescript-service-template/blob/main/.github/workflows/ci.yml for example, the env is provided to the `run` call that actually does the install, and the `npmrc` echo uses single quotes, so it puts the placeholder in the npmrc, not the actual secret

mwvolo: moved it down as a script variable

TomWoodward (approved):

PR #2954, opened 2026-08-02
Bump flex-page-renderer to renderer-1.1.22-rc.4 (dev test RC) — supersedes #2953

mwvolo:
## Summary
Points `@openstax/flex-page-renderer` at **`renderer-1.1.22-rc.4`** for dev review. **Supersedes #2953** (`renderer-1.1.22-rc.3`) — a strict superset.

**New since rc.3** (flex-pages [#27](https://github.com/openstax/flex-pages/pull/27)):
- **Big Numbers block** — a new container block that holds Big Number child blocks and lays them out in a reflowing "impact-stats" row. The child slot accepts only Big Number blocks.
- **Well `layout` removed** — the Well no longer doubles as a row-flow container (per review); that use case now lives in the Big Numbers block.
- **Well `heading_style` help reworded** (clarity only).

⚠️ Pairs with **openstax-cms [#1765](https://github.com/openstax/openstax-cms/pull/1765)** (remove Well layout, reword heading_style help, add the `big_numbers` block + migration).

**Carried over:** QuoteBlock fix (rc.3), tables/container-queries/hero/tabs/accordion (rc.1–2), plus the CMS pairings below.

## CMS pairings
| CMS PR | For | Status |
|---|---|---|
| #1756 | per-card colors | merged |
| #1761 | accordion per-item colors | open |
| #1764 | quote image optional | merged |
| #1765 | Well layout removal + Big Numbers block | open |

## Test
- `test/src/pages/flex-page.test.tsx` — 18/18 against rc.4.
- `yarn lint` clean.
- Recommend a dev deploy to eyeball the new Big Numbers block and confirm the Well no longer offers a layout option.

Supersedes #2953 — closeable once reviewed.

Issue #27, opened 2016-02-12
Scoped stylesheets do not allow for @keyframes and @font-face in Firefox


melaridi:
I found a bug reported here:  [(https://bugzilla.mozilla.org/show_bug.cgi?id=830056)]

PR #2964, opened 2026-08-12
Bump flex-page-renderer to renderer-1.1.25-rc.1 (dev test RC) — supersedes #2954

mwvolo:
Supersedes #2954. Same fixes, pointed at a newer RC that also carries the table cell slot.

## The RC

`renderer-1.1.25-rc.1` is cut from an integration of **both** open flex-pages PRs:

- `core-2403-mobile-and-components` (#27) — mobile responsiveness + componentized CSS overrides
- `table-cell-slot` (#34) — the per-cell render slot

`1.1.25` sits above #27's `1.1.24` so there's no lineage confusion with the existing `1.1.22-rc.*` tags, which came from #27's line only. Cut with `scripts/publish.bash` following the established pattern (bump the renderer, pin internal refs to it, subtree-split, tag).

Nothing is merged in flex-pages yet — this RC exists so dev can exercise both PRs together. It goes stale if either moves.

## Carried over from #2954

- `raw-html.tsx` — effect dependency fix, so embedded scripts don't re-execute on re-renders with unchanged markup
- `block-map.ts` — `html` block override, so JavaScript authored in an html block actually runs (`innerHTML` never executes `<script>`)
- `flex-page.test.tsx` — the `cardsBlock` per-card colour assertions, updated for the shape change in flex-pages #30 (`data.accentColor` → `findByType(data.config, 'accent_color')`). This is why that test failed against ≥1.1.20; it was a stale fixture, not a renderer regression.

## New here: instructor-resource links in tables

Flex tables can be filled from CMS book-resource data. The CMS cannot emit URLs for access-locked instructor resources — it renders at page-serialize time with no request user, behind a CDN with no cookie variance — so it emits a book-page fallback link plus a `resource_ref` marker.

This adds a `table` override that resolves those markers client-side, where verified-instructor status is visible, and renders them through the renderer's new `TableCellContext` slot.

The cell renders the **real resource-box button** (`LeftContent`), not a substituted URL. That matters: the Give dialog and the `trackLink` call that posts to `/salesforce/download-tracking/` — creating the user-behavior adoption record — only happen on the dialog's own anchor. Rewriting cell data alone would have produced a bare `<a>` and silently dropped every adoption record.

Everything else the table does — sorting, filtering, striping, row colours, row limits, anchor ids — stays with the renderer's own component; the override only claims marked cells and returns `undefined` for the rest. Tables with no markers render the delegate with the original data and fire no fetch.

Loading and unmatched cells fall through to the CMS fallback link rather than blanking. Note `books/resources/` is faculty-only server-side, so a `Student` marker currently always resolves to unmatched and keeps its fallback.

## Also included

A dedicated `toggle` / `toggle-control-bar` test (also up separately as #2963). Those files had no direct tests and were covered only incidentally through consumers, via a `throttle(…, 125)` — so coverage flipped on timing and failed the repo-wide 100% threshold intermittently on `main`.

## Verification

Against the real `renderer-1.1.25-rc.1` install:

```
npx tsc --skipLibCheck --noEmit   → clean
npx eslint src                    → clean
npx jest ./test/src               → exit 0, 183 suites, 772 tests
```

One caveat worth stating: this repo still has a residual coverage flake unrelated to anything here — an occasional `99.98%/99.98%/99.95%` with no branch failure, which I saw once and could not reproduce across repeated clean runs. Distinct from the toggle one, which this branch fixes.


mwvolo:
Superseded by #2965, which pins the real `renderer-1.1.24` release tag now that flex-pages #27 and #34 are both merged to main. Every change from this branch carried over — the diff against main is identical apart from the tag ref.

